PR #4, opened 2024-06-21
Add sample workflow to test ci vm runner

ttmchiou:
PR to show working CI workflow and successful CI integration

Please rename the workflow when integration work begins.
Suggest to finish porting GH Orgs first

### DIFF
--- a/.github/workflows/hello-world.yaml
+++ b/.github/workflows/hello-world.yaml
@@ -15,5 +15,5 @@ jobs:
   hello-world:
     runs-on: ["in-service", "n150"]
     steps:
-      - name: Hello World
-        run: echo "Hello World!"    
+      - name: Sample Action
+        run: echo "Action run!"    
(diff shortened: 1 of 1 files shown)


.github/workflows/hello-world.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+permissions:
+  actions: read
+  contents: read
+  pages: write
+  id-token: write
+
+jobs:
+  hello-world:
+    runs-on: ["in-service", "n150"]
```
ayerofieiev-tt, line 16: what does "in-service" means?

ayerofieiev-tt (approved):

.github/workflows/hello-world.yaml
ttmchiou: the label in-service, just means that the runner is in service.
Often times we need to do maintainence so we use this label to enable/disable the runner from executing CI jobs

PR #20, opened 2024-07-03
Add conversion for aten.transpose

mcw-zwakeelTT:
### Problem description
Transformation for torch transpose op to ttnn equivalent.

### What's changed
Added conversion routine in to_tt_pass.py
Added tests for transpose in a separate test file "test_ops.py"


tests/test_ops.py
```diff
@@ -0,0 +1,60 @@
...
+        input = torch.randint(0, 3, input_shapes[0]).type(torch.bfloat16)
+        dim0 = input_shapes[1]
+        dim1 = input_shapes[2]
+        result_before = m.forward(input, dim0, dim1)
+        option = torch_ttnn.TorchTtnnOption(device=self.device)
+        option.gen_graphviz = True
+        # The compilation is lazy, so we need to run forward once to trigger the compilation
+        m = torch.compile(m, backend=torch_ttnn.backend, options=option)
+        result_after = m.forward(input, dim0, dim1)
+        option._out_fx_graphs[0].print_tabular()
```
ayerofieiev-tt, line 43: do we need this print?

tests/test_ops.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # Check the graph has be rewritten and contain ttnn ops
+        nodes = list(option._out_fx_graphs[0].nodes)
+        for node in nodes:
+            print(node.name)
```
ayerofieiev-tt, line 48: do we need these?

torch_ttnn/passes/to_tt_pass.py
```diff
@@ -368,6 +368,17 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
                     new_node,
                     delete_user_cb=lambda node: node != new_node,
                 )
+            if node.target == torch.ops.aten.transpose.int:
```
ayerofieiev-tt, line 371: what `int` stands for in `transpose.int`?

ayerofieiev-tt, line 371: Is there an in_place version of this operation?

torch_ttnn/passes/to_tt_pass.py
```diff
@@ -368,6 +368,17 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
...
+                dim0 = args[1]
+                dim1 = args[2]
+                rank = len(node.meta["val"].size())
+                permutation = list(range(rank))
+                permutation[dim0], permutation[dim1] = permutation[dim1], permutation[dim0]
+                new_node = g.call_function(ttnn.permute, args=(args[0], permutation))
+                node.replace_all_uses_with(
+                    new_node,
+                    delete_user_cb=lambda node: node != new_node,
```
ayerofieiev-tt, line 380: what does this do?

tests/test_ops.py
ayerofieiev-tt, line 1: Whats the logic behind a name of this file? 
Lets call it `test_transpose.py` and keep it for a single op?

tests/test_ops.py
mcw-zwakeelTT: This printing is there for all existing tests, so I included it.

tests/test_ops.py
mcw-zwakeelTT: This we can remove, I forgot to remove at the last moment.

torch_ttnn/passes/to_tt_pass.py
mcw-zwakeelTT: Integer inputs. There could be other overloaded variants of transpose. This one is for integers.

tests/test_ops.py
mcw-zwakeelTT: test_more_ops.py contains tests for all existing ops. Thought the file was getting lengthy, so kept a separate file called it test_ops.py. We can merge it to have one single test python file, or if we want we can have separate files for individual ops. That way we can keep adding more variants of tests for a single op in future. But the downside could be too many files for tests.

torch_ttnn/passes/to_tt_pass.py
mcw-zwakeelTT: This replaces the users of existing torch node attaching to the users list of newly created ttnn node. Then the torch node can be deleted safely.

torch_ttnn/passes/to_tt_pass.py
mcw-zwakeelTT: Not that I am aware of.

tests/test_ops.py
kevinwuTT: Yeah all the tests have this from the start. I think it should be fine since it only shows up if we ask pytest to be verbose.

tests/test_ops.py
kevinwuTT: Yeah let's start separating ops into individual test files. There will be some repeated boiler-plate code, but it won't be too much. I will start refactoring the entire repo soon and reorganizing the tests will be part of it.

mcw-zwakeelTT: Okay from now on, I will write separate tests for individual ops.

torch_ttnn/passes/to_tt_pass.py
ayerofieiev-tt: how hard is it to handle non-integer transposes? why don't we just cover all variants here?

tests/test_ops.py
ayerofieiev-tt: Please rename this file in this PR. Thanks

ayerofieiev-tt:
@mcw-zwakeelTT you can see that checks failed - thats because PR contains files which were not formatted and thats likely because you did not install pre-commit.

mcw-zwakeelTT:
> @mcw-zwakeelTT you can see that checks failed - thats because PR contains files which were not formatted and thats likely because you did not install pre-commit.

I have added a new commit covering this and renamed the test file.

torch_ttnn/passes/to_tt_pass.py
mcw-zwakeelTT: It can be covered, have to check how many overloads are there and include them. I will capture this in the next commit.

mcw-zwakeelTT: @ayerofieiev-tt I had a look at transpose overloads. There are two of them:
**torch.ops.aten.transpose.int** (uses integer indices to specify the dimensions to be transposed)
**torch.ops.aten.transpose.Dimname** (uses dimension names to specify the dimensions to be transposed)
Dimname is used for tensor dimension having names rather than integer indices. So int overload pretty much covers all input datatypes.

kevinwuTT (approved):

PR #21, opened 2024-07-03
Implement conversion for aten.expand by using ttnn.repeat

kevinwuTT:
### Problem description
Transform aten.expand by using ttnn.repeat. 

### What's changed
The interpretation of the `shape` argument between the two ops differ. For aten.expand the `shape` argument is the desired output shape. For ttnn.repeat the `shape` argument is a multiplier of the input shape instead. Therefore the shape information of the input tensor is needed.
* Add retention of shape metadata information for Placeholder nodes since transform() removes them
  * Add missing metadata for conversions done under ReplaceMoreTtManually
* Implement the conversion from aten.expand using the shape metadata
 
In addition, ttnn.repeat acts differently between TILE and ROW_MAJOR layouts. In most cases, we would need to change the layout of a node to ROW_MAJOR before passing that to ttnn.repeat. Similarly, we would need to check if a layout change back to TILE is needed if any following node that uses ttnn.repeat is a tt compute node.
* Add logic to insert layout changes before and/or after ttnn.repeat

### DIFF
--- a/torch_ttnn/passes/to_tt_pass.py
+++ b/torch_ttnn/passes/to_tt_pass.py
@@ -7,6 +7,7 @@ from ..utils import (
     DummyDevice,
     DummyTtnnTileLayout,
 )
+import numpy as np
 
 try:
     import ttnn
@@ -167,6 +168,7 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
                 new_node = g.call_function(
                     ttnn.clone, args=(args[0], memory_config, dummy_dtype)
                 )
+                new_node.meta["val"] = node.meta["val"]
                 node.replace_all_uses_with(
                     new_node,
                     delete_user_cb=lambda node: node != new_node,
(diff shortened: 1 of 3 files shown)


torch_ttnn/passes/to_tt_pass.py
```diff
@@ -364,20 +375,50 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
...
+
+        # Save Placeholder metadata before calling transform()
+        input_nodes = {
+            node.name: node.meta for node in gm.graph.nodes if node.op == "placeholder"
+        }
+
         # Replace more patterns with torch.fx.Transformer
         gm = ReplaceMoreTt(gm).transform()
 
+        # Restore metadata for Placeholder nodes
```
ayerofieiev-tt, line 417: why do we lose metadata?

kevinwuTT: From my investigations so far, it could be an oversight. I found some very recent discussion https://discuss.pytorch.org/t/understanding-torch-fx-traceback-preserve-node-meta/205319 with the same issue but unfortunately no response yet. I could create an issue to track this actually. Could be something we optimize away at a later time if we find a better solution.

kevinwuTT: https://github.com/tenstorrent-metal/pytorch2.0_ttnn/issues/22

PR #23, opened 2024-07-03
Add conversion for aten.t.default

mcw-zwakeelTT:
### Problem description
Conversion pass for aten.t.default to its ttnn equivalent; ttnn.permute in this case. aten.t.default is a subset of transpose and can be realized with permute op.

### Proposed
Covers 3 types of input.
- 2D tensor: transpose via exchanging both dims using ttnn.permute
- 0D (scalar) & 1D tensor: Returned as it is. Right now no conversion takes place for this type of input.
- Any other N-D tensor: Throws an assert.

### Clarification
Is there a identity ttnn op which takes a tensor and returns it as it is?


tests/test_aten_t.py
```diff
@@ -0,0 +1,122 @@
+import torch
+import torch_ttnn
+import unittest
+from torch_ttnn import ttnn
```
ayerofieiev-tt, line 4: can we do simply `import ttnn`?

mcw-zwakeelTT: Yeah that can be done as well.

kevinwuTT: Just FYI, mock_ttnn will be removed during the refactor very soon.

kevinwuTT (approved):

kevinwuTT:
@mcw-zwakeelTT We have some new directories, so let's put this under: `tests/lowering/tensor_manipulation`. We'll fix test_transpose later. Also don't forget the `import ttnn`.

PR #26, opened 2024-07-05
Reorganize files

ayerofieiev-tt:
### Ticket
Link to Github Issue

### Problem description

### What's changed


tests/lowering/conv/test_conv.py
kevinwuTT, line 1: This one is weird because originally it's meant to test tracing with the torch_stat backend that outputs a .csv file and other visual graphics. There definitely should be a conv test but probably when the lowering gets implemented.

kevinwuTT (approved):

PR #34, opened 2024-07-09
Add conversion for aten.unsqueeze

mcw-zwakeelTT:
### Problem description
Conversion pass for aten.unsqueeze to its ttnn equivalent; ttnn.reshape in this case.

### Constraints
ttnn.reshape requires input tensor to be in **row major** order.
ttnn.reshape can be computed **on device** only if input tensor is to be reshaped to **4D**
Unsqueezing the last dimension throws a runtime error (buffer is not allocated). Bert doesn't have such unsqueeze operations, so test_bert is passing.


tests/lowering/tensor_manipulation/test_unsqueeze.py
```diff
@@ -0,0 +1,131 @@
...
+
+class Unsqueeze3Module(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, x, y):
+        return torch.unsqueeze(x, y)
+
+    def input_shapes(self):
+        return [(5, 2, 4, 3), -2]
```
ayerofieiev-tt, line 42: I think this should be a single module. Input shapes should not be inside. This will help to eliminate duplication

mcw-zwakeelTT: I tried with a single module creating class objects from one class module, but for some reason it errors out. @kevinwuTT also faced similar issues earlier. Multiple modules exist for previous tests, we can keep this for now.

ayerofieiev-tt (approved):

PR #43, opened 2024-07-16
Polish metrics table in README

kevinwuTT:
- Add links to the model name that points to the directory of the test
- Changed run success to use emojis
- Add unique torch op counts
- Change precision of numeric metrics to 2 decimal places
- Fix spacings
- Add a note about measuring multiple graphs

### DIFF
--- a/tools/collect_metrics.py
+++ b/tools/collect_metrics.py
@@ -20,16 +20,16 @@ csv_header_mappings = {
         "Run Success",
         "Indicates whether the model runs successfully after conversion.",
     ),
-    "torch_ops_before": (
-        "Torch Ops (Before)",
-        "The number of operations used by the model in the original Torch implementation.",
+    "torch_ops_total_unique_before": (
+        "Torch Ops Before (Unique Ops)",
+        "The total number of operations used by the model in the original Torch implementation. The number in parenthesis represents the total unique ops.",
     ),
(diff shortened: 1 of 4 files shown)


tools/collect_metrics.py
```diff
@@ -46,6 +46,18 @@ def load_pickle(path: str):
...
+model_link_mappings = {
+    "BERT": "tests/models/bert",
+    "Bloom": "tests/models/bloom",
+    "Falcon": "tests/models/falcon",
+    "GPT-2": "tests/models/gpt2",
+    "Llama": "tests/models/llama",
+    "Mnist (Eval)": "tests/models/mnist",
+    "Mnist (Train)": "tests/models/mnist",
+    "ResNet18": "tests/models/resnet",
+    "YOLOS": "tests/models/yolos",
```
ayerofieiev-tt, line 58: I'd love us to get rid of this over time. It should be enough to just add a test, there should be no need to "register it"

ayerofieiev-tt (approved):

tools/collect_metrics.py
kevinwuTT: Yeah this could be reworked. My rational is that sometimes the model name and associated directory are different, so I used a mapping to link them. But we could also have this data generated from the test model itself as another way.

PR #50, opened 2024-07-23
Preserve intermediate node meta for composite ops

mcw-zwakeelTT:
### Brief

- Restore node meta for intermediate nodes created during lowering of composite ops
- Add checks in composite op tests

### DIFF
--- a/tests/lowering/eltwise/binary/test_div.py
+++ b/tests/lowering/eltwise/binary/test_div.py
@@ -35,6 +35,10 @@ def test_div(device, input_shapes):
     assert target.count(ttnn.mul) == 1
     assert target.index(ttnn.reciprocal) < target.index(ttnn.mul)
     assert nodes[target.index(ttnn.mul)].args[1].target == ttnn.reciprocal
+    # Intermediate node meta check if preserved
+    for node in nodes:
+        if node.target == ttnn.full or node.target == ttnn.reciprocal:
+            assert node.meta["val"].size() == input_shapes[0]
     # Check inference result
     assert check_with_pcc(result_before, result_after)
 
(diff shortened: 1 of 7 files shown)


tests/lowering/eltwise/binary/test_div.py
```diff
@@ -35,6 +35,10 @@ def test_div(device, input_shapes):
     assert target.count(ttnn.mul) == 1
     assert target.index(ttnn.reciprocal) < target.index(ttnn.mul)
     assert nodes[target.index(ttnn.mul)].args[1].target == ttnn.reciprocal
+    # Intermediate node meta check if preserved
+    for node in nodes:
```
ayerofieiev-tt, line 39: Any idea how this can be added centrally so one don't have to check it like this in every test?

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -235,6 +239,10 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
                     full_node = g.call_function(
                         ttnn.full, args=(arg_metadata.size(),), kwargs=new_kwargs
                     )
+                    # Intermediate node meta is not preserved, this ensures retention
+                    # Are output dims same for full node and actual node to be replaced?
```
ayerofieiev-tt, line 243: I think output dims are not guaranteed to be the same due to tiling

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -266,15 +274,22 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
                 # out = beta * input + alpha * (batch1 @ batch2)
                 # if beta is 0, input is ignored, and nan and inf in it will not be propogated
                 new_node = g.call_function(ttnn.matmul, args=(args[1], args[2]))
+                # Intermediate node meta is not preserved, this ensures retention
+                new_node.meta["val"] = node.meta["val"]
```
ayerofieiev-tt, line 278: Any ideas on how to make this a common routine for conversions to avoid duplication and make sure it is consistently used across all conversions?

ayerofieiev-tt (approved):
I don't want to block this, but please think how to make such things a part of the infrastructure, so that each individual conversion don't have to care about it that much

torch_ttnn/passes/lowering/to_tt_pass.py
mcw-zwakeelTT: I think the meta field stores non-tiled output shape for each of these.

PR #52, opened 2024-07-30
Simplify testing of models to minimize extra code

kevinwuTT:
* Add Pytest autouse fixture that compiles a model and records metrics for original and compiled run
  * Models no longer need to import ttnn or torch_ttnn modules
  * Models will need to register a few lines of information for this fixture to work
* Simplify the collect metrics script
  * No longer need to hard-code a map of model paths
  * No longer need to import output artifacts to calculate accuracy
    * Accuracy is calculated in the pytest fixture directly

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,13 @@
 import pytest
 import ttnn
 import torch
+import torch_ttnn
+import collections
+from tests.utils import calculate_accuracy
+import time
+from pathlib import Path
+import os
+import pickle
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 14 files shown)


ayerofieiev-tt (commented):
Goooood sign! 
Screenshot 2024-07-30 at 1 07 16 PM

tests/conftest.py
```diff
@@ -15,3 +22,61 @@ def reset_torch_dynamo():
     # PyTorch caches models. Start a fresh compile for each parameter of the test case.
     torch._dynamo.reset()
     yield
+
+
+@pytest.fixture(autouse=True)
```
ayerofieiev-tt, line 27: does autouse apply it to lowering tests too?

ayerofieiev-tt: Checked on a call. Yes. 
We agreed it is ok for now. 
We agreed this will go away during the next step.

ayerofieiev-tt (approved):

PR #53, opened 2024-08-09
Fix model breakages

kevinwuTT:
# Model status for this PR:
Screenshot 2024-08-12 at 6 07 26 PM

## ⚠️ GPT-2 fixes
[Unmark gpt2 and mnist test models to expect passing](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/0c3d249d6439521a63a3be76b7f0a3b66d30aafd)
[Disable conversion from aten._to_copy](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/4d3143b00596e824f746dbdc581ad3f5e10d02b2)
~~[Pass device for all from_torch ops](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/6903996c10c2d6eef35910a943ccf5360c03e825)~~ Reverted because this conflicts with unsqueeze conversion
[Replace aten.full op to a literal scalar for certain cases](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/8cd16f25b13452679c3008a6fe09a3d684e11a09)
[Compare only Tensor types for dictionary outputs](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/12d6326d521df2fdf7220f33ed317bf5f05d8f5b)
Fails because:
* Explicit casting of dtype to ttnn.bfloat16 for `from_torch` op makes GPT-2 fail. Removing it causes, Bloom, Llama, and Yolos to fail.
```
Failed to generate binaries for embeddings_tilize TT_THROW @ ../tt_metal/jit_build/build.cpp:396: tt::exception
info:
ncrisc build failed
```

## MNIST fixes:
[Replace aten.view with aten.reshape](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/9ebb604dca54255e27e179a402e7a98735411ddc)

## ⚠️ Falcon-7B status:
Fails because:
* layer_norm conversion
```
TT_THROW @ ../tt_metal/impl/program/program.cpp:489: tt::exception
info:
Statically allocated circular buffers on core range [(x=0,y=0) - (x=0,y=0)] grow to 1611072 B which is beyond max L1 size of 1048576 B
```

## Bloom and Llama fixes:
[Add conversion for aten.min](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/1b5f5ae34232e8952a2c0413167c288bae5755c7)
[Add exception to aten.eq conversion](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/f846f2493393d0195f84ecaeaf931afe53ad2526)
[Fix reusing ttnn data movement op if mixed with aten ops](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/c0625bc154b20b8b50fa986336d8f620880784e9)
[Convert all inputs to ttnn.bfloat16 when moving data in](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/80ddac100585c1dd14c439048bf9114dfef41aa3)
[Skip unsqueeze transformation if last dim of input is not the same as…](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/83d6c692876e2a1d4efd6805dc0538737378b89e)
[Add exception to aten.expand conversion when last dimension of input …](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/c4c998e9c0688cbd70ea108d1036f8cf1c76d530)
[Support list type arguments](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/a2e096f05550706331f5ade4aabceca1887bad1f)
[Check layout change for ttnn reshape and embedding op](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/e4fadb28fb8d8194bc87bb0f1c397b31667cd483)
[Freeze encoder for llama model](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/f4a9e2f4acdf4af9aadbe642a0011d96bcd9df7f)

## Yolos fixes
[Add workaround for ttnn.permute when dim 0 is 1 for rank 3](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/2a0e588c69fb83b0652e5d872e600dec5cfcea34)
[Reconvert int64 types from metadata when mixing ttnn and aten ops](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/d10d715092f5e680e6b866ccf739e8ac6e2973d7)
[Check for valid page size for ops that decompose to ttnn.full](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/53b753782bdd3c259e4de8708e1ffcf3424c36a9)
[Delete aten.expand op if output has the exact same shape](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/a3c7dc8cc22ed6b6c55541718e490d5b91220cbb)

## General fixes:
[Consolidate metadata during op conversion](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/919d390bc7776f914030d7644e7e09d22524bf43)
[Fix output type of aten.arange unit test to match output of original](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/3950c9c63cc518eac480dfb68e824539d4f4f5ed)
[Disable to_copy unit test to re-evaluate conversion](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/f093dd7b88a7048f7c7cef64b9889cca1a88ee1f)
[Lower pcc for addmm slightly](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/30474068fa6c59a61d3273522afeb151def79cda)
[Change input shapes of some unit test to match exceptions in current …](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/e7f1936e28f2877861535b8e145424144d15fdf7)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,17 +8,17 @@ This project allows to run PyTorch code on [Tenstorrent](https://tenstorrent.com
 
 The table below summarizes the results of running various ML models through our TTNN compiler. For each model, we track whether the run was successful, the number of operations before and after conversion, the number of `to_device` and `from_device` operations, performance metrics, and accuracy.
 
-| Model                               | Run Success   | Torch Ops Before (Unique Ops)   | Torch Ops Remain (Unique Ops)   | To/From Device Ops   |   Original Run Time (ms) | Compiled Run Time (ms)   | Accuracy (%)   |
-|:------------------------------------|:--------------|:--------------------------------|:--------------------------------|:---------------------|-------------------------:|:-------------------------|:---------------|
-| [Mnist (Eval)](tests/models/mnist)  | ✘             | 14 (8)                          | 5 (4)                           | 16                   |                    36.12 | N/A                      | N/A            |
-| [Mnist (Train)](tests/models/mnist) | ✅            | 14 (8)                          | 7 (5)                           | 14                   |                   114.49 | 2742.8                   | 81.75          |
-| [ResNet18](tests/models/resnet)     | ✅            | 70 (9)                          | 42 (4)                          | 47                   |                  2094.6  | 10950.18                 | 99.99          |
-| [Bloom](tests/models/bloom)         | ✘             | 1407 (29)                       | N/A                             | N/A                  |                  9127.68 | N/A                      | N/A            |
-| [YOLOS](tests/models/yolos)         | ✘             | 964 (28)                        | N/A                             | N/A                  |                  1353.22 | N/A                      | N/A            |
-| [Llama](tests/models/llama)         | ✘             | 3 (3)                           | 1 (1)                           | 5                    |                 52926.3  | N/A                      | N/A            |
-| [BERT](tests/models/bert)           | ✅            | 1393 (21)                       | 537 (4)                         | 1607                 |                 65342    | 61028.65                 | 98.64          |
-| [Falcon](tests/models/falcon)       | ✘             | 3 (3)                           | 1 (1)                           | 5                    |                 47738.8  | N/A                      | N/A            |
-| [GPT-2](tests/models/gpt2)          | ✘             | 748 (31)                        | N/A                             | N/A                  |                  2287.61 | N/A                      | N/A            |
+| Model                               | Run Success   | Torch Ops Before (Unique Ops)   | Torch Ops Remain (Unique Ops)   |   To/From Device Ops |   Original Run Time (ms) | Compiled Run Time (ms)   | Accuracy (%)   |
+|:------------------------------------|:--------------|:--------------------------------|:--------------------------------|---------------------:|-------------------------:|:-------------------------|:---------------|
+| [Mnist (Eval)](tests/models/mnist)  | ✅            | 14 (8)                          | 5 (4)                           |                   16 |                    36.62 | 477.12                   | 98.33          |
+| [Mnist (Train)](tests/models/mnist) | ✅            | 14 (8)                          | 7 (5)                           |                   14 |                   117.66 | 2571.71                  | 68.54          |
+| [ResNet18](tests/models/resnet)     | ✅            | 70 (9)                          | 42 (4)                          |                   47 |                  2025.6  | 10350.92                 | 99.99          |
+| [Bloom](tests/models/bloom)         | ✅            | 1407 (29)                       | 627 (12)                        |                 1375 |                 28876.2  | 65763.95                 | 45.77          |
+| [YOLOS](tests/models/yolos)         | ✅            | 964 (28)                        | 409 (11)                        |                  919 |                  1516.37 | 45219.57                 | 71.71          |
+| [Llama](tests/models/llama)         | ✅            | 5 (4)                           | 3 (2)                           |                    3 |                168160    | 186989.25                | 34.41          |
+| [BERT](tests/models/bert)           | ✅            | 1393 (21)                       | 564 (7)                         |                 1488 |                 64262    | 57553.72                 | 98.64          |
+| [Falcon](tests/models/falcon)       | ✘             | 3 (3)                           | 2 (2)                           |                    5 |                 39306.6  | N/A                      | N/A            |
+| [GPT-2](tests/models/gpt2)          | ✘             | 748 (31)                        | 330 (15)                        |                  530 |                  5889.45 | N/A                      | N/A            |
 
 ### Explanation of Metrics
 
@@ -47,7 +47,7 @@ The table below summarizes the results of running various ML models through our 
 | aten.max_pool2d_with_indices.default | ✘        |       1 |
 | aten.relu.default                    | ✅       |       3 |
 | aten.t.default                       | ✅       |       2 |
-| aten.view.default                    | ✘        |       1 |
+| aten.view.default                    | ✅       |       1 |
 #### Mnist (Train)
 | aten ops                             | status   |   count |
 |:-------------------------------------|:---------|--------:|
@@ -58,7 +58,7 @@ The table below summarizes the results of running various ML models through our 
 | aten.native_dropout.default          | ✘        |       2 |
 | aten.relu.default                    | ✅       |       3 |
 | aten.t.default                       | ✅       |       2 |
-| aten.view.default                    | ✘        |       1 |
+| aten.view.default                    | ✅       |       1 |
 #### ResNet18
 | aten ops                                          | status   |   count |
 |:--------------------------------------------------|:---------|--------:|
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -70,43 +70,141 @@ The table below summarizes the results of running various ML models through our
...
-| aten.div.Tensor                | ✅       |      24 |
+| aten.div.Tensor                | ✘        |      24 |
 | aten.embedding.default         | ✅       |       3 |
 | aten.expand.default            | ✅       |      96 |
 | aten.gelu.default              | ✅       |      24 |
 | aten.mul.Tensor                | ✅       |       1 |
 | aten.native_layer_norm.default | ✅       |      49 |
 | aten.permute.default           | ✅       |      96 |
-| aten.rsub.Scalar               | ✅       |       1 |
+| aten.rsub.Scalar               | ✘        |       1 |
```
ayerofieiev-tt, line 160: really can't support?

tests/models/gpt2/test_gpt2.py
```diff
@@ -17,11 +17,20 @@ def test_gpt2(record_property):
...
         model_name, torch_dtype=torch.bfloat16
     )
+    for param in m.parameters():
+        param.requires_grad = False
     m.eval()
 
     # Set up sample input
     test_input = "This is a sample text from "
-    inputs = tokenizer(test_input, return_tensors="pt")
+    inputs = tokenizer.encode_plus(
```
ayerofieiev-tt, line 26: can you clarify what happens here?

tests/models/gpt2/test_gpt2.py
```diff
@@ -17,11 +17,20 @@ def test_gpt2(record_property):
     m = AutoModelForSequenceClassification.from_pretrained(
         model_name, torch_dtype=torch.bfloat16
     )
+    for param in m.parameters():
+        param.requires_grad = False
```
ayerofieiev-tt, line 21: why?

kevinwuTT: I noticed GPT-2 is using backward ops like `torch.ops.aten.native_layer_norm_backward.default`. Both backward and forward version return a tuple of 3 outputs, but the backwards version will use all 3 instead of just the first one. So for simplicity, I wanted to disable backward ops since we haven't added support for them yet. I could remove this and test with default parameters. Or maybe parameterize this and test with and without grad.

tests/models/gpt2/test_gpt2.py
kevinwuTT: In this case, `encode_plus` isn't any different since I'm not using the extra information like masking that `encode_plus` provides. I believe BERT was because the input had multiple sentences. I can revert this until the changes make sense.

README.md
kevinwuTT: Thanks for noticing this. I fixed it with a better check for valid page sizes https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/commits/a6aa0c2f9744d203b673ba41f36ea9f348d2cbe2. rsub and div should be converted for BERT again.

tests/models/gpt2/test_gpt2.py
ayerofieiev-tt: Such things ideally should be done by infra

kevinwuTT: That's fair. Most of the current transformer models share a lot of commonalities, so that can be merged to maybe a pytest fixture to reduce duplicates. Since GPT-2 isn't passing at this point, I can remove these from this PR for now.

tests/models/gpt2/test_gpt2.py
kevinwuTT: Since GPT-2 isn't working in this PR anyway, I reverted this change.

tests/models/gpt2/test_gpt2.py
kevinwuTT: This change to GPT-2 has been reverted.

tests/models/llama/test_llama.py
```diff
@@ -14,12 +13,22 @@ def test_llama(record_property):
     tokenizer = AutoTokenizer.from_pretrained(
         model_name, padding_side="left", torch_dtype=torch.bfloat16
     )
+    tokenizer.pad_token = tokenizer.eos_token
```
ayerofieiev-tt, line 16: I bet this can be removed

tests/lowering/tensor_manipulation/test_unsqueeze.py
```diff
@@ -64,7 +65,8 @@ def test_unsqueeze2(device, input_shape, dim):
 
 @pytest.mark.parametrize(
     "input_shape, dim",
-    [((5, 2, 4, 3), -2)],
+    # [((5, 2, 4, 3), -2)],
```
ayerofieiev-tt, line 68: can you leave a comment in the code on why this is commented and not removed and what should be done to uncomment? ideally with a link to github issue

tests/lowering/tensor_manipulation/test_unsqueeze.py
```diff
@@ -17,7 +17,8 @@ def forward(self, x, y):
 
 @pytest.mark.parametrize(
     "input_shape, dim",
-    [((5, 2, 4, 3), 1)],
+    # [((5, 2, 4, 3), 1)],
```
ayerofieiev-tt, line 20: same as https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53/files#r1717556225

tests/lowering/creation/test_to_copy.py
```diff
@@ -21,6 +21,9 @@ def forward(self, x):
         return torch.add(to, to)
 
 
+# aten.to_copy is used to convert a dtype to another.
```
ayerofieiev-tt, line 24: btw, I think we got a ttnn.typecast to convert

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -32,91 +34,102 @@ def are_args_from_int_output_ops(args):
...
+            call_func = self.call_function_prop_meta(target, args, kwargs)
 
-        # Copy metadata of old node to replacement
-        meta = fx_traceback.get_current_meta()
-        if meta is not None and "val" in meta:
-            call_func.node.meta["val"] = meta["val"]
+        # # Copy metadata of old node to replacement
+        # meta = fx_traceback.get_current_meta()
+        # if meta is not None:
+        #     call_func.node.meta = meta
```
ayerofieiev-tt, line 132: remove?

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -264,109 +280,101 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
...
+                    for node_user in node_users:
+                        if node_user.target == torch.ops.aten.div.Tensor:
+                            node_user.update_arg(1, args[1])
             if node.target == torch.ops.aten.baddbmm.default:
                 kwargs = node.kwargs
                 # out = beta * input + alpha * (batch1 @ batch2)
                 # if beta is 0, input is ignored, and nan and inf in it will not be propogated
                 new_node = g.call_function(ttnn.matmul, args=(args[1], args[2]))
-                # Intermediate node meta is not preserved, this ensures retention
-                new_node.meta["val"] = node.meta["val"]
```
ayerofieiev-tt, line 278: I thought we have to carry over meta["val], no?

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -264,109 +280,101 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
...
+                            kwargs=new_kwargs,
+                        )
+                        to_layout = g.call_function(
+                            ttnn.to_layout, (full,), {"layout": TtnnTileLayout()}
+                        )
+                        recip = g.call_function(ttnn.reciprocal, (to_layout,), {})
+                    else:
+                        recip = g.call_function(ttnn.reciprocal, (args[1],), {})
+                    new_node = g.call_function(ttnn.mul, (args[0], recip), {})
+                    new_nodes.append(new_node)
```
ayerofieiev-tt, line 367: Not a matter of this PR. But want to leave thoughts here.

I think of multiple types of conversions:
1. **Trivial conversion** is a conversion which allows full 1-1 mapping. In this case it is just a record in a map
2. **Arg Order conversion** is a conversion which maps close to 1-1 but requires arguments reordering
3. **Args Change conversion** is a conversion which maps close to 1-1 but requires arguments to be modified
4. **Lowering conversion** is a conversion which maps 1 aten into multiple TT-NN ops

I don't like seeing a long file for 400 lines of code with all conversions at the same place.
Ideally conversions are separate, stored in some registry, and generic conversion mechanism just pulls the required conversion.

ayerofieiev-tt (approved):

torch_ttnn/passes/lowering/to_tt_pass.py
kevinwuTT: So same idea as https://github.com/tenstorrent/pytorch2.0_ttnn/pull/53#discussion_r1717567654. `g.call_function` will now carry over the metadata of the current node automatically.

tests/models/llama/test_llama.py
kevinwuTT: Unfortunately, it seems like if I want to pad the inputs, Llama will emit this error if I don't specify a pad_token. 

```
ValueError: Asking to pad but the tokenizer does not have a padding token. Please select a token to use as `pad_token` `(tokenizer.pad_token = tokenizer.eos_token e.g.)` or add a new pad token via `tokenizer.add_special_tokens({'pad_token': '[PA...
```

tests/lowering/tensor_manipulation/test_unsqueeze.py
kevinwuTT: A [commit](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/6903996c10c2d6eef35910a943ccf5360c03e825) I made for GPT-2 affected this test, but I reverted it since GPT-2 isn't working entirely.

PR #62, opened 2024-08-21
Fix models GPT-2 and Falcon-7B breakages

kevinwuTT:
Fixes GPT-2:
- [Skip explicit type cast to bfloat16 when moving from_torch for certain ops](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/c5abd3c8e0b01db9dcda39b9d390ae753322cc03)
- [Do reshape on device if rank is 2 and 0th dim is 1](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/c3d601b3fec5874dea51666298f1190587ffb3fa)
- [Redo validating page size for TILE and ROW_MAJOR layouts](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/a4976449e1fb6c79cf511da94119ddff898f95ea)

Fixes Falcon-7B:
- Other strategies like input splitting would work better. Deferred fixing Falcon-7B for now.
- ~~[Reduce the default `hidden_size` of 4544 because this cannot fit `ttnn.layer_norm`](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/62/commits/a260708b2323cec356c2b386ef95ea34339602b8)~~
- ~~[Restrict lowering to ttnn.arange if steps are negative](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/62/commits/0101b8ac3c7ec96846a01bd2046c868385ea2744)~~


tests/models/falcon/test_falcon.py
```diff
@@ -1,25 +1,37 @@
...
-@pytest.mark.xfail
 def test_falcon(record_property):
     record_property("model_name", "Falcon")
 
     # Download model from cloud
     model_name = "tiiuae/falcon-7b-instruct"
+
+    # Falcon hidden_size is 4544, which is too big to fit into layer_norm
+    config = AutoConfig.from_pretrained(model_name)
+    config.hidden_size = config.hidden_size // 2
```
ayerofieiev-tt, line 15: I think we should avoid such changes to models.
We should target to intake a model with 0 modifications. 

Think that any modification is a "fine".

If model does not fit - ideally compiler figures that input should be split.

@boris-drazic @mcw-zwakeelTT here is an example of input and op which require splitting.

ayerofieiev-tt: If we are to modify a model like that, lets at least always keep the original model with 0 modifications. 
Our target workflow - user takes a model, adds compile and it just runs

kevinwuTT: Since is the default value for this pretrained model, I'll revert this and mark it as xfail again. I have an issue https://github.com/tenstorrent/pytorch2.0_ttnn/issues/63 to track and make it easier to test this scenario.

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -83,6 +83,16 @@ def is_tt_data_move(node) -> bool:
     ]
 
 
+# For non-tiled shapes, these ttnn ops require layout changes
+layout_change_ops = set(
```
ayerofieiev-tt, line 87: Alright, so we need to query if an op supports non-tilized tensors so that this can be done in a better way?

kevinwuTT:
Closing at outdated. Both models appear to be running on main, although not end-to-end.

PR #72, opened 2024-08-23
Test Whisper

jdh8:
### Ticket
None

### Problem description
The test fails for now
```
self = <OpOverload(op='aten.convolution', overload='default')>
args = (TorchTensor([[[-0.0413, -0.0265,  0.0159,  ..., -0.0378, -0.0378,  0.0947],
              [-0.0312, -0.0400, -0.0216,...0.4448, -0.3418, -0.7529, -0.0972,  0.0148,  0.2161,  5.0508,  0.8599],
       requires_grad=True), [2], [1], [1], ...)
kwargs = {}

    def __call__(self, *args, **kwargs):
>       return self._op(*args, **(kwargs or {}))
E       RuntimeError: expected scalar type BFloat16 but found Float

../tt-metal/python_env/lib/python3.8/site-packages/torch/_ops.py:513: RuntimeError
```

### What's changed
Add test for Whisper
https://huggingface.co/openai/whisper-small


tests/models/whisper/test_whisper.py
```diff
@@ -0,0 +1,24 @@
...
+    model = WhisperForConditionalGeneration.from_pretrained("openai/whisper-small")
+    model.config.forced_decoder_ids = None
+
+    # load dummy dataset and read audio files
+    ds = load_dataset("hf-internal-testing/librispeech_asr_dummy", "clean", split="validation")
+    sample = ds[0]["audio"]
+    input_features = processor(sample["array"], sampling_rate=sample["sampling_rate"], return_tensors="pt").input_features 
+
+    # generate token ids
+    predicted_ids = model.generate(input_features)
```
kevinwuTT, line 19: From my understanding, `.generate` does extra processing to the outputs compared to calling `model(input_features)`. Can you confirm the outputs between these two methods just on CPU? The torch_ttnn backend calls the compiled `model(input_features)`, so there might be some mismatching issues. I know `.generate` is a very popular way of using models, so that might be something we can try to support in the near future.

tests/models/whisper/test_whisper.py
```diff
@@ -0,0 +1,26 @@
+from transformers import WhisperProcessor, WhisperForConditionalGeneration
+from datasets import load_dataset
+
+
```
kevinwuTT, line 4: If compilation is xfail include this `@pytest.mark.compilation_xfail`

ayerofieiev-tt (approved):

PR #77, opened 2024-08-27
Add torch_only marker to mark compiled run as xfail but leave torch run as normal

kevinwuTT:
When we introduce a model test, we want to ensure that the torch/CPU version runs. The compiled version may be expected to fail at this time. The `torch_only` mark will raise an `pytest.xfail` instead of the normal exception when compiled run fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,12 +297,14 @@ You can also substitute the backend with `torch_stat` to run a reference compari
 # Add a model test
 If you want to record run time metrics for a model or test, include a Pytest fixture named `record_property` as a parameter and set the "model_name" key.  
 If you also want to compile the model with torch_ttnn backend, set the "torch_ttnn" key to a tuple in this order `(model, test_inputs, outputs)`. "model_name" still needs to be set. See the example code snippet below. Currently, only `torch.nn.Module` models with a `forward` function are supported.
-```
+```python
 def Model(torch.nn.Module):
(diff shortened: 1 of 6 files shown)


ayerofieiev-tt (approved):

README.md
```diff
@@ -297,12 +297,14 @@ You can also substitute the backend with `torch_stat` to run a reference compari
...
 If you also want to compile the model with torch_ttnn backend, set the "torch_ttnn" key to a tuple in this order `(model, test_inputs, outputs)`. "model_name" still needs to be set. See the example code snippet below. Currently, only `torch.nn.Module` models with a `forward` function are supported.
-```
+```python
 def Model(torch.nn.Module):
     def forward(self, x):
         # ...
         return outputs
 
+# Add torch_only marker if torch/CPU runs, but compiled version is xfail
+@pytest.mark.torch_only
```
ayerofieiev-tt, line 307: @kevinwuTT lets find a better name.

This should indicate that we know that this fails run on tenstorrent hardware and we are okay with it for now.


Maybe `compile_fail_expected`?

kevinwuTT: @ayerofieiev-tt `compilation_xfail` to correspond to the native `pytest.xfail` maybe?

ayerofieiev-tt: @kevinwuTT , sounds good

PR #119, opened 2024-08-29
Model OPT.

cthsieh:
### Ticket
None

### Problem description
Add model OPT to trace.

### What's changed
Add model OPT.


tests/models/opt/test_opt.py
```diff
@@ -0,0 +1,29 @@
...
+
+    model_inputs = tokenizer([prompt], return_tensors="pt")
+
+    model_inputs["max_new_tokens"] = 30
+    model_inputs["do_sample"] = False
+    generated_ids = model.generate(**model_inputs)
+
+    tokenizer.batch_decode(generated_ids)[0]
+
+    record_property("torch_ttnn", (model, model_inputs, generated_ids))
```
ayerofieiev-tt, line 29: this should be `model.generate` as model.generate(..) is used on line 25

ayerofieiev-tt (changes requested):
please pass `model.generate` to torch_ttnn property

tests/models/opt/test_opt.py
cthsieh: Absolutely. Thanks for the catch.

ayerofieiev-tt (approved):

PR #127, opened 2024-08-29
Model Autoencoder (conv)

ayerofieiev-tt:
### Ticket
None

### Problem description
This was merged and now reverted as it caused issues.
Please try to run locally
CC @cthsieh 

Add model Autoencoder (conv) to trace.

Solved by writing a new test.

### What's changed
Add model Autoencoder (conv).


cthsieh:
We resolved the issue by manually crafting a convolutional autoencoder model instead of using one from a third-party package.

The third-party package, called "diffusers," is popular and widely used on Hugging Face, with hundreds of projects relying on it. However, the package primarily supports model inference through a "pipeline", which is not compatible with the Torch Dynamo compiler flow.

We've kept the original test but marked it as skipped because using the pipeline is still a valid approach, and we might need to support it in the future. Additionally, we created a new test that bypasses the pipeline, ensuring compatibility with our current compiler flow.

PR #129, opened 2024-08-30
Model segment-anything.

cthsieh:
### Ticket
None

### Problem description
Add model segment-anything to trace. 

However, unfortunately the goal is not achieved. We decided to push the code just for keeping a record, and we marked the test skipped.

The test fails to run because it requires installation of package sam2, but sam2 requires Python>=3.10.0 while the default version on Ubuntu 20.04 is 3.8. 

We tried to find alternative pytorch implementation of the same model but found nothing.

### What's changed
Add model segment-anything.


ayerofieiev-tt:
Got it. No problem. I hope we will move to 22.04 soon and it should enable this test. Thank you

ayerofieiev-tt (approved):

PR #135, opened 2024-08-30
Add additional sections to ProblemSolving doc

kevinwuTT:
### Ticket
None

### Problem description
Provide support for some common error messages 

### What's changed
* Add a section to cover errors hidden by Pytorch when a failure happens from the backend
* Add a section to cover incompatible function arguments

### DIFF
--- a/docs/ProblemSolving.md
+++ b/docs/ProblemSolving.md
@@ -17,7 +17,7 @@ If you see an error in the test output like this:
 ```
 ERROR tests/models/bert/test_bert.py::test_bert - TypeError: __call__(): incompatible function arguments. The following argument types are supported:
 ```
-### Read he Problem
+### Read the Problem
 Examine the log above the error message. It includes detailed information, for example:
 ```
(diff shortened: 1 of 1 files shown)


docs/ProblemSolving.md
```diff
@@ -89,3 +89,88 @@ if tt_tensors_device.layout == ttnn.TILE_LAYOUT:
...
+>>> import pdb
+>>> pdb.post_mortem(e.__traceback__)
+```
+***
+```
+> /home/ubuntu/repo/tt-metal/build/python_env/lib/python3.8/site-packages/torch/fx/node.py(98)_get_qualified_name()
+-> raise RuntimeError("Unable to represent lambda") from e
+```
+
+## Incompatible Function Arguments
```
ayerofieiev-tt, line 146: > Incompatible Function Arguments

In this way we now have 2 sections with the same name, right?
Lets distinguish between wrong arguments passed (like position vs kw vs missing vs type mismatch).

This particular one is on a type mismatch, but more specifically for Tensor type mismatch, which requires a special handling.

ayerofieiev-tt (approved):

docs/ProblemSolving.md
kevinwuTT: Done. Renamed to `Mismatch Between Torch and TTNN Tensor Types`

PR #138, opened 2024-09-01
Model YOLOv3.

cthsieh:
### Ticket
None

### Problem description
Add model YOLOv3 to trace.

Deriving a PyTorch implementation of the YOLOv3 model can be quite challenging. YOLOv3 was NOT originally developed in PyTorch, so it relies on contributions from the community for porting. Many of these ported versions have issues, such as not working as claimed, lacking pretrained weights, or depending on complex dependencies like `libGL.so`. 

To address these challenges, we studied the tt buda demos and adopted the same approach to write YOLOv3 test: Described the model topology (see directory `holli_src`) ourselves and sourced pretrained weights from external links.

### What's changed
Add model test YOLOv3.


tests/models/yolov3/test_yolov3.py
```diff
@@ -0,0 +1,42 @@
...
+
+from tests.models.yolov3.holli_src.yolov3 import *
+
+
+@pytest.mark.compilation_xfail
+def test_yolov3(record_property):
+    record_property("model_name", "YOLOv3")
+
+    # Download model weights
+    url = "https://www.ollihuotari.com/data/yolov3_pytorch/yolov3_coco_01.h5"
```
ayerofieiev-tt, line 17: this tests downloaded neat the test .py file?

cthsieh: You are right. That's not good. So I modified to download the model weights into the system `/tmp` directory.

Besides, I added to check the existence of the downloaded model weights. If the weight has been downloaded before and has not been cleared by the system, it can be reused by the tests, which saves time and bandwidth to download again, while weight is usually large.

ayerofieiev-tt: We carefully clean cache between model test runs to make sure we fit into the allocated disk space. 
Lets please put it inside `~/.cache/custom_weights`
I will make sure to clean this in the workflow.

cthsieh: It makes sense. Changed as you described.

PR #142, opened 2024-09-01
Model Unet-carvana.

cthsieh:
### Ticket
None

### Problem description
Add model Unet-carvana to trace.

While many of the model tests extract models from popular hubs such as HuggingFace, TorchVision, and Torch Hub, this test retrieves a model directly from GitHub. To facilitate this process, I wrote a utility function called `github_download`. I suggest moving this function into `conftest.py` in the future when this test is proven to work charmingly.

Of course I have tried running this test locally without problem.

### What's changed
Add model Unet-carvana.


ayerofieiev-tt (changes requested):
Lets avoid this pass please. In this case there are 2 options. Skip this model or copy it.
Lets avoid such downloads.

cthsieh:
@ayerofieiev-tt  
This PR has passed the workflow: https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/10846415409

I have removed the GitHub downloads and rewritten with a self-contained model description. You can merge it if you feel comfortable with the change.

ayerofieiev-tt (approved):

PR #148, opened 2024-09-04
Convert `aten._unsafe_view` like `aten.view`

jdh8:
### Ticket
None

### Problem description
Lower `aten._unsafe_view`

### What's changed
Convert `aten._unsafe_view` to `ttnn.reshape`

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -250,7 +250,7 @@ class ReplaceMoreTt(torch.fx.Transformer):
         if target == torch.ops.aten.permute.default:
             return self.call_function_prop_meta(ttnn.permute, args, kwargs)
 
-        if target == torch.ops.aten.view.default:
+        if target in (torch.ops.aten.view.default, torch.ops.aten._unsafe_view.default):
             # aten.reshape is more stable if the input nodes have changed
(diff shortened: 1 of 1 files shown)


kevinwuTT:
Is there a test for this? I wonder what lowers into _unsafe.view.

jdh8:
I have no idea either.  The only thing I know is:
```
// _unsafe_view() differs from view() in that the returned tensor isn't treated
// as a view for the purposes of automatic differentiation. (It's not listed in
// VIEW_FUNCTIONS in gen_autograd.py).  It's only safe to use if the `self` tensor
// is temporary. For example, the viewed tensor here (a + b) is discarded immediately
// after viewing:
//
//  res = at::_unsafe_view(a + b, size);
```
https://github.com/pytorch/pytorch/blob/af64789cfa3aef37926fb3d9601d5e47406098c8/aten/src/ATen/native/TensorShape.cpp#L959-L965

Treating it as a normal `view` should be safe all the time since we don't apply autodiff.

jdh8:
Already taken care by c0a948e4787e266085771e372b96017ad0768ec9

https://github.com/tenstorrent/pytorch2.0_ttnn/blob/196ae752825cab6c532fa8ed741f3acd030385a3/torch_ttnn/passes/lowering/to_tt_pass.py#L700-L701

PR #168, opened 2024-09-06
Show conversion status of each input variation in docs

kevinwuTT:
### Ticket
None

### Problem description
We want to know the conversion status of an input variation. This requires mapping metadata of the original aten op and using this information to determine if an op with a specific input variation has been converted. However, when using the `Transformer` class during lowering, the metadata for the inputs are not migrated for the newly transformed graph. However, this class maintains a copy of the old graph that can be used to perform this migration.

### What's changed
- [x] For each ttnn op, save the original aten op metadata during lowering
  - [x] Migrate the metadata for the graph inputs, i.e. Placeholder nodes
- [x] Map the aten op input variation with the corresponding ttnn op
- [x] Generate the tables with the new conversion status columns
- [x] Cleanup in `collect_metrics.py`
- [x] Cleanup in `metrics.py`
- [x] Merge main and upload latest metrics


tools/collect_metrics.py
```diff
@@ -79,99 +87,17 @@ class ConversionStatus(Enum):
...
-
-
 def serialize_schema_metrics_to_operations(metrics):
     """Transform schema information to a list of `class Operation` pydantic models."""
     operations = []
     for node in metrics:
-        op_schema_string = stringify_schema_to_string(node["schema"], node["input_shapes"], node["input_values"])
-        operations.append(pydantic_models.Operation(op_name=node["opname"], op_schema=op_schema_string))
+        if "original_inputs" in node:
+            # This might not be right
```
ayerofieiev-tt, line 95: seems fine, why it might not be right?

ayerofieiev-tt (approved):

tools/collect_metrics.py
kevinwuTT: This is currently fine. The TTNN ops store the schema for the original aten ops in "original_inputs", so I wanted to remind myself to double check this. I think later if we want to save inputs for TTNN ops, we might slightly change this part.

PR #140, opened 2024-09-01
Draft: Model Hand Landmark.

cthsieh:
### Ticket
None

### Problem description
Add model Hand Landmark to trace.

This test depends on fixture manage_depedencies which was developed in PR 137. So, review PR 137 first, and then I continue to finish this PR in case of any change according to PR 137.

### What's changed
Add model Hand Landmark.

PR #141, opened 2024-09-01
Model Pose Landmark.

cthsieh:
### Ticket
None

### Problem description
Add model Pose Landmark to trace.

This test depends on fixture `manage_depedencies` which was developed in PR 137. So, review PR 137 first, and then I continue to finish this PR in case of any change according to PR 137.

### What's changed
* Add model Pose Landmark.
* Expand README.md to describe the use of Git LFS.

PR #171, opened 2024-09-09
Model Hand Landmark, with Git LFS.

cthsieh:
### Ticket
None.

### Problem description
This is a revision for PR #140. The binary files are now committed by Git LFS, considering that they are binary and we don't care about their history.

### What's changed
Add test for model Hand Landmark, including a test image (in the context of a "hand", of course) and a model task object file.


ayerofieiev-tt:
@cthsieh which file is not commited with LFS?

cthsieh:
> @cthsieh which file is not commited with LFS?

Files involved in this PR.
```
hand_landmark
├── hand_landmarker.task            LFS file.
├── test_hand_landmark.py          Normal git file.
└── woman_hands.jpg                  LFS file.
```

ayerofieiev-tt (approved):

cthsieh:
The CI failed due to the same reason as in #141, which needs `git lfs pull` to derive the LFS files.

https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/10860030910/job/30140837416

ayerofieiev-tt:
I added git lfs pull to the workflow. Need to check why one model passed and another did not

cthsieh:
Passed the CI. https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/10875361939

The trick is to make the test perform `git lfs pull` by itself.

PR #178, opened 2024-09-11
Convert `aten.fill.Scalar` to `ttnn.full`

jerrysky3:
### Ticket
None

### Problem description
Convert `aten.fill.Scalar` to `ttnn.full`. It first generates a tensor aligned to 32x32 in `TILE_LAYOUT` then crop the results into the required shape. For tile layout unsupported case (dims < 2 and > 4), it fallbacks to `ROW_MAJOR_LAYOUT`

### What's changed
- Add conversion from `aten.fill.Scalar` to `ttnn.full`
- Add a new fx.wrap function `target_wrappers.crop` to crop tensor into the target shape
- Test cases for the conversion

### DIFF
--- a/tests/lowering/eltwise/unary/test_fill.py
+++ b/tests/lowering/eltwise/unary/test_fill.py
@@ -13,23 +13,23 @@ class FillScalarModule(torch.nn.Module):
 
 
 @pytest.mark.parametrize(
-    "input_shape, value, converted",
+    "input_shape, value",
     [
-        ((32, 32), 0.5, True),
-        ((32, 1), 114, True),
-        ((1, 32), 120, True),
-        ((4, 4), 32, True),
-        ((1, 2, 100, 100), 16, True),
-        ((1, 1, 1, 1), 61, True),
-        ((2, 1, 1, 1), 64, True),
-        ((1, 1, 2, 1), 8, True),
-        ((1, 1, 1, 2, 2), 1.0, True),
-        ((64,), 1.0, True),
-        # Unsupported: scalar
-        ((), 1.0, False),
+        ((32, 32), 0.5),
+        ((32, 1), 114),
+        ((1, 32), 120),
+        ((4, 4), 32),
+        ((1, 2, 100, 100), 16),
+        ((1, 1, 1, 1), 61),
+        ((2, 1, 1, 1), 64),
+        ((1, 1, 2, 1), 8),
+        ((1, 1, 1, 2, 2), 1.0),
+        ((64,), 1.0),
+        # TODO(#256): Not support scalar tensor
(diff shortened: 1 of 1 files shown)


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -318,9 +318,11 @@ def torch_dtype_to_ttnn_dtype(dtype: torch.dtype):
...
 # For valid non-interleaved buffers page size 2048 must equal buffer size X. For interleaved-buffers page size should be divisible by buffer size
 def has_valid_page_size(shape, strict=False):
-    if len(shape) >= 2 and shape[-1] > 0:
-        return shape[-1] % 32 == 0 or (not strict and shape[-1] < 32)
-    return False
+    if len(shape) < 2 or shape[-1] == 0:
+        return False
+    if not strict and shape[-1] < 32:
+        return True
+    return shape[-2] % 32 == 0 and shape[-1] % 32 == 0
```
jerrysky3, line 325: I found that the previous strict condition is still not enough. The buffer size needs to be divisible by or equal to the page size 2048 by this assert https://github.com/tenstorrent/tt-metal/blob/7040c7836d2c2953971834981ff79ff21c0d36e5/tt_metal/impl/buffers/buffer.cpp#L33-L38

And even though `(16, 64)` bfloat16 tensor, which size = 2048, will pass assertions, it gives out wrong results (in `test_gt.py` tests) or memory corruputions (in tests add by this PR) during my testing. I assume it is due to [the only supported tile shape is 32x32](https://github.com/tenstorrent/tt-metal/blob/main/tech_reports/tensor_layouts/tensor_layouts.md#321-tile-shapes) currently?

Therefore, I add the condition here to check both dims must be divisible by 32.

tests/lowering/eltwise/unary/test_fill.py
```diff
@@ -0,0 +1,46 @@
...
+
+    option = torch_ttnn.TorchTtnnOption(device=device, gen_graphviz=True)
+    # The compilation is lazy, so we need to run forward once to trigger the compilation
+    m = torch.compile(m, backend=torch_ttnn.backend, options=option)
+    result_after = m.forward(input, value)
+    option._out_fx_graphs[0].print_tabular()
+
+    # Check the graph has be rewritten and contains ttnn ops
+    nodes = list(option._out_fx_graphs[0].nodes)
+    # Currently only tensor sizes divisible by 32x32 are converted.
```
jerrysky3, line 40: It seems that there are multiple places currently have this limitiation and skip the conversion for now. Should I create an issue to track these?

torch_ttnn/passes/lowering/to_tt_pass.py
jerrysky3: It now supports non-tiled size

jerrysky3:
It can support non-aligned size now

PR #179, opened 2024-09-11
Convert `aten.slice.Tensor` to `ttnn.slice`

jdh8:
### Ticket
None

### Problem description
- PyTorch lowers `input[:, :, start:end]` to 3 `aten.slice.Tensor` ops for each dimensions.  I'd like to eliminate the outer two no-op slicing ops, but I need help from @kevinwuTT.
- Note that ending indices of `ttnn.slice` are inclusive

### What's changed
Convert `aten.slice.Tensor` to `ttnn.slice` in supported cases

PR #180, opened 2024-09-11
Refactor manual lowering

jdh8:
### Ticket
None

### Problem description
Improve readability of `ReplaceMoreTtManually`

### What's changed
The list `new_nodes` was a local variable.  Its writing sides scattered hundreds of lines, but it was only read once.  I refactored generation of `new_nodes` to a function.  Writing `new_nodes` becomes returning a new node, which boosts readability.  We can also reuse the RAUW logic by simply returning a node.

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -354,8 +354,8 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
                 arg_metadata = node.meta["val"]
                 ttnn_dtype = torch_dtype_to_ttnn_dtype(arg_metadata.dtype)
                 # Add additional logic to choose the appropriate memory_config type: DRAM or L1
-                new_node = g.call_function(target_wrappers.clone, args=(args[0],))
-                return new_node
+                return g.call_function(target_wrappers.clone, args=(args[0],))
+
             if node.target == torch.ops.aten.native_layer_norm.default:
                 new_node = g.call_function(
                     ttnn.layer_norm,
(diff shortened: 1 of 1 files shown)


jdh8:
This refactoring also makes eliding no-ops easier, which helps converting `aten.slice.Tensor` (#179).  Besides, our use cases of `aten._adaptive_avg_pool2d` look like no-op as well:

> ### aten._adaptive_avg_pool2d.default
> |    | ATen Input Variations                                                | Status   |
> |---:|:---------------------------------------------------------------------|:---------|
> |  0 | Tensor<[1, 512, 7, 7]> self = ?,<br>List[int]<> output_size = [7, 7] | Done     |
>
> -- https://github.com/tenstorrent/pytorch2.0_ttnn/blob/main/docs/cumulative_input_variations.md#aten_adaptive_avg_pool2ddefault

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -346,17 +346,17 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
     nodes = list(gm.graph.nodes)
     for node in nodes:
         g = GraphWrapper(node)
-        args = node.args
 
-        new_nodes = []
-        with g.inserting_before(node):
-            # TODO (kevinwuTT): consolidate and simplify these statements?
+        def rewrite_node(node):
```
ayerofieiev-tt, line 350: I remember in some cases this conversion created >1 node. Do you return a list?

jdh8: No, we only need the last node, the node that replaces the old node.  The current code uses only the last node as well.

https://github.com/tenstorrent/pytorch2.0_ttnn/blob/a5ce2a677f6ea9c6ba704c21db1c771bff03ef8a/torch_ttnn/passes/lowering/to_tt_pass.py#L592-L596

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -367,30 +367,33 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
                 node_users = list(new_node.users.keys())
                 for node_user in node_users:
                     node_user.replace_all_uses_with(new_node)
```
jdh8, line 369: In this particular case, we need a manual RAUW loop.  This piece of code remains the same.

torch_ttnn/passes/lowering/to_tt_pass.py
jdh8: For example, if the old graph is
```
Y = gemm(A, B, C)
```

And we convert it to a new graph
```
X = A @ B
Y = X + C
```

We only need to replace uses of the `gemm` node to the `add` node.  The intermediate `matmul` is good on its own (and Python takes care of its lifespan).

kevinwuTT (approved):
This looks good. Removes a lot of redundancy for new conversions.

PR #223, opened 2024-09-19
Convert `aten.alias` to no-op

jerrysky3:
### Ticket
No

### Problem description
Convert `aten.alias` to no-op

### What's changed
- Replace `aten.alias` with its input (no-op)
- Add test cases


jdh8 (approved):
I verified this changeset with `pytest tests/lowering/`, but I could not `pytest tests/models/` due to insufficient disk space.

Issue #249, opened 2024-09-20
Generate tests with all input variations


swimdi:
As this message from @ayerofieiev-tt , want to generated the input variations test and mark xFAIL
```
I have a question/suggestion
We have traces for all ops
https://github.com/tenstorrent/pytorch2.0_ttnn/blob/main/docs/cumulative_input_variations.md#atenrepeatdefault
We know all input variations used across models
https://github.com/tenstorrent/pytorch2.0_ttnn/pull/189/files#diff-d9d8d4b791253df05409adb7ef6dcf69bce6939960dbd7d593bf3aa859b0775aR19
But when we define tests we do not rely on that data yet
So there is a lot of back and forth.
We run lowering test - it works on the simpest case
Then when we run models tests we find that something does not work and it takes time to extract which shape/op it is
Suggestion.
* Lets add all known input variations to the op test.
* If some input fails
  * mark xfailed
  * handle in the lowering to fallback for that case
We currently do the same in TT-NN - we use traces to Sweep ops. We then take a model and see which ops do fallback and improve generality to start running e2e.
I assume it is an easy change of approach which can make the overall process more efficient (edited) 
Wdyt?
(I ask you, assuming you are the team lead)
```

PR #250, opened 2024-09-20
Generate tests/input_variations

swimdi:
### Ticket
https://github.com/tenstorrent/pytorch2.0_ttnn/issues/249
### Problem description
To implement the script of generate input variations test

### What's changed

As the ticket says, it want to make all input varations as tests from [the input variations doc](https://github.com/tenstorrent/pytorch2.0_ttnn/blob/main/docs/cumulative_input_variations.md#atenrepeatdefault)

I study the code of collect-metrics job, seems it use `InputVarPerOp::write_md_for_input_variations` to generate the input variations doc, so I think use `InputVarPerOp` to collect all model metrics, and its recorded input variations will consistent with the input variations doc, so I reuse these data structure of `InputVarPerOp`.

Currently, the test flow of generated input variation test is:

0. convert input variation string as randomized value
1. inference before compiled (native inference)
2. compile with ttnn backend
3. inference after compiled
4. check ttnn op is in compiled graph
5. check accuracy between before/after inference

And as my observation, every stage will have someone who makes a mistake, for example

0. convert input variation string as randomized value
 - for `aten.add.Tensor`, one of its input variation is `['Tensor<[s0 + 1, s0 + 1]> self = ?', 'Tensor<> other = 16']`, which `s0` is unknown from the input variations doc, so cannot randomize input value. These type of case already mark as SKIP

1. inference before compiled (native inference)
Should every test pass this stage, but not
 - for `aten._unsafe_index.Tensor`, one of its input variation is `['Tensor<[1, 64, 15, 20]> self = ?', "List[Optional[Tensor]]<> indices = [None, None, 'torch.Size([30, 1])', 'torch.Size([40])']"]`, and will cause` IndexError: index 30 is out of bounds for dimension 0 with size 15`,  not sure why these illegal input shape appear in the input variations doc

4. check ttnn op is in compiled graph
If the target aten op of this input variation test is not support of conversion to ttnn, then this stage will failed

I'm not sure what the PASS standard of these input variation test is
 - for example, `aten._native_batch_norm_legit_functional.default` will not convert to ttnn op, but it can successfullly inference and pass accuracy, so it should not cause model broken, but should it mark as pass or failed?

So currently I'm not add the input variation test into CI.

But I add the job of "generate" input variation test into CI, it will be triggered after model-tests job, and will make `tests/input_variation/` as downloadable artifacts.

btw, current pytest option of input variation test has:
 - pytest --input_var_check_ttnn: check has ttnn op if enable
 - pytest --input_var_check_accu: check accuracy if enable
 - pytest --input_var_only_native: only inference before compiled if enable

c.c. @cthsieh

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,6 +1,7 @@
 import torch
 import numpy as np
 import collections
+import re
 
 
 class ModelTester:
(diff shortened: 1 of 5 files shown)


tests/utils.py
```diff
@@ -252,9 +253,129 @@ def calculate_accuracy(original_outputs, compiled_outputs):
...
+        accuracy = torch.mean(torch.tensor(output_pccs)).item()
     elif isinstance(original_outputs, torch.Tensor) and isinstance(compiled_outputs, torch.Tensor):
         # Handle case where outputs are Pytorch Tensors
         _, accuracy = comp_pcc(original_outputs, compiled_outputs)
     else:
         accuracy = None
     return accuracy
+
+
+def get_input_vals_from_metric_str(op_name, input_strings):
```
ayerofieiev-tt, line 281: this method itself requires good testing

tools/generate_input_variation_test_from_models.py
```diff
@@ -0,0 +1,101 @@
...
+        def _unjoin_br(str_br: str):
+            return str_br.split(",<br>")
+
+        sort_by_opname = dict(sorted(self.items()))
+        basedir.mkdir(parents=True, exist_ok=True)
+        for opname, inputs_variations in sort_by_opname.items():
+            inputs_strings = [_unjoin_br(input_variations) for input_variations in inputs_variations.keys()]
+            opname_ = opname.replace(".", "_")
+            filename = f"test_{opname_}.py"
+            text = f"""
```
ayerofieiev-tt, line 18: can we move this to a separate file, smth like `aten_test_template`?

tests/utils.py
swimdi: Ok, Can I add the testing after this PR merged?

ayerofieiev-tt (approved):

tools/generate_input_variation_test_from_models.py
swimdi: Had moved to `tools/aten_test.tmpl`

tools/generate_input_variation_test_from_models.py
```diff
@@ -0,0 +1,101 @@
+import os
+from pathlib import Path
+from tools.collect_metrics import load_pickle, InputVarPerOp
+
+
+class InputVarTestExporter(InputVarPerOp):
```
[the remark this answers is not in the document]
swimdi: Had renamed as `AtenOpTestExporter`

PR #255, opened 2024-09-22
Enhance mnist training test.

cthsieh:
### Ticket
None

### Problem description
After reviewing `test_mnist.py` as an example of testing the training mode, I noticed that it lacks a test for the backward pass, which calculates the gradients. I've rewritten the test, and if you agree with this change, I plan to apply the same approach to all other training tests.

### What's changed
* The training test in `test_mnist.py`.
* The training process in `conftest.py`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -83,6 +83,25 @@ def compile_and_run(device, reset_torch_dynamo, request):
 
     if "torch_ttnn" in record:
         model, inputs, outputs = record["torch_ttnn"]
+
+        try:
+            # Why `inputs.requires_grad`?
+            #
+            # Backward pass computes gradients for all trainable weight tensors.
+            # However, verifying every gradient can be costly, especially for
+            # large models with many parameters.
+            #
+            # Instead of checking each gradient, we check the gradient
+            # of the "model input" tensor only. Computing the input gradient
+            # serves as an indicator for the health of all other gradients.
+            # Based on the "chain rule", the input gradient depends on all other
+            # gradients, so any incorrect gradient computation should reflect here.
+            is_backward_checked = True if model.training and inputs.requires_grad else False
+        except Exception as e:
+            # Exception to handle unsupported case of multiple model inputs.
+            # Currently, we assume the model has a single input for backward pass checks.
+            is_backward_checked = False
+
         try:
             # Compile model with ttnn backend
             option = torch_ttnn.TorchTtnnOption(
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -83,6 +83,25 @@ def compile_and_run(device, reset_torch_dynamo, request):
 
     if "torch_ttnn" in record:
         model, inputs, outputs = record["torch_ttnn"]
+
+        try:
```
ayerofieiev-tt, line 87: I am not clear why you need this try/catch block, it does not call any functions and I can’t see how it can throw.

ayerofieiev-tt: Is this for the case when inputs or model do not contain a key?

tests/models/mnist/test_mnist.py
```diff
@@ -37,6 +37,14 @@ def forward(self, x):
 def test_mnist_train(record_property):
     record_property("model_name", "Mnist (Train)")
 
+    # Fixing the random seed for reproducibility to ease debugging.
+    #
+    # Training processes involve more randomness compared to evaluation,
+    # such as random initialization of weights.
+    # Setting a fixed random seed is crucial for consistent testing
+    # and debugging during the training process.
+    torch.manual_seed(99)
```
ayerofieiev-tt, line 46: 👍

tests/conftest.py
kevinwuTT: Yeah, at least `requires_grad` is an attribute of `torch.Tensor` [type](https://pytorch.org/docs/stable/generated/torch.Tensor.requires_grad_.html). I've seen other models have other types for their inputs (see below), so I think this PR only supports `Tensor` input types at the moment. 

https://github.com/tenstorrent/pytorch2.0_ttnn/blob/6f78bd3a6497a8bb5246e3d46fce431cadc50c37/tests/conftest.py#L172-L178

ayerofieiev-tt (approved):

kevinwuTT:
I think this looks good. Just a note, we will need to support the new backwards ops and most of them will return a tuple of outputs.

Screenshot 2024-09-23 at 11 13 23 AM

kevinwuTT (approved):

tests/conftest.py
cthsieh: The try/catch is to handle the fact that `inputs.requires_grad` is invalid if `inputs` is an instance of type `collections.Mapping` or `collections.Sequence`.

In other words, we assume that the models under train test must have only one input.

Will relax the assumption if necessary.

PR #261, opened 2024-09-24
Add lowering of split to enable BERT

boris-drazic:
### Ticket
n/a

### Problem description
Lower torch split to tt split for BERT model.

### What's changed
Added lowering on split to to_tt_pass

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -1,5 +1,6 @@
 import torch
 import ttnn
+import math
 from torch_ttnn.utils import (
     GraphCleanup,
     TtnnBfloat16,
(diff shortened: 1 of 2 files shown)


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -660,6 +661,24 @@ def rewrite_node(node):
                 else:
                     # Fallback: aten.reshape is more stable if the input nodes have changed
                     return g.call_function(torch.ops.aten.reshape.default, args, kwargs)
+            if node.target == torch.ops.aten.split.Tensor:
```
ayerofieiev-tt, line 664: @sjameelTT can you please take a look at this?
The expectation is that we can simply call ttnn.split here but you see we have to make rm input, calculate an arg, change an arg. 

Please review, maybe something can be removed and hopefully everything can get hidden in the op itself.

ayerofieiev-tt (approved):
We can go with this to unblock, but need to improve `ttnn.split` because we can't go far with this approach

PR #265, opened 2024-09-25
Enable SqueezeBert

boris-drazic:
### Ticket
n/a

### Problem description
SqueezeBert model test is labeled as xfail, it should be enabled.

### What's changed
Tokenization of input to model has changed so that input are padded to 256 and model weights are loaded as bfloat16.
As a result, embedding layer receives inputs that are bfloat16 and multiple of tile size on last dimension ... and the model passes the test.

### DIFF
--- a/tests/models/squeeze_bert/test_squeeze_bert.py
+++ b/tests/models/squeeze_bert/test_squeeze_bert.py
@@ -5,14 +5,21 @@ import torch
 import pytest
 
 
-@pytest.mark.compilation_xfail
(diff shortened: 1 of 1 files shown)


ayerofieiev-tt:
@boris-drazic , should close this now?

boris-drazic:
Closing without merging as PR https://github.com/tenstorrent/pytorch2.0_ttnn/pull/267 enabled squeeze bert by adding fallback for embedding.

Issue #269, opened 2024-09-27
Refactor model test


swimdi:
Current model test have a torch_ttnn model run if set torch_ttnn property, but as the training mode join, this run_model function become insufficient
```
# in training mode it should need to run backward to collect the backward ops
def run_model(model, inputs):
    if isinstance(inputs, collections.Mapping):
        return model(**inputs)
    elif isinstance(inputs, collections.Sequence):
        return model(*inputs)
    else:
        return model(inputs)
```

So there propose to refactor the model test, let it have a `ModelTester` class and handle the run model flow

PR #274, opened 2024-09-30
Apply model tester

swimdi:
### Ticket
#269 

### Problem description
Continue with #271 , apply `ModelTester` class to all the `tests/models` test

### What's changed
 - Apply `ModelTaster` class to all the `tests/models` test
 - [This comment](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/271#discussion_r1779753157) mention the convert bf16 problem, and after I doing apply, I think the better site to convert bf16 is in the `_load_model` & `_load_inputs` stage, so I remove convert bf16 code from `set_model_*` & `set_input_*`
 - Remove old usage of `record_property("torch_ttnn", (model, input, output)`

This PR needs to merge #271 first


ayerofieiev-tt:
@swimdi one concern about this change is that out intension with model tests was to try to keep models in as pristine state as possible. We wanted users to see from tests that they can take a model and just run it by just adding a single line to compile it. 

I ask to keep this into consideration. I assume its not possible to sit on both chairs and assume that the change you made is the only way to achieve the desired outcome (trace both inference and training via common infrastructure). But I want to check with you, in case you have ideas on how to minimize impact on models even more.

ayerofieiev-tt (approved):

swimdi:
@ayerofieiev-tt  Or if I just let the testcase implement `test_model_eval` & `test_model_train` function, and the testcase is generated from template, which can just register detail function like `set_model_eval` / `append_fake_loss_function`, would this modify better?

```
# tests/models/test_xxx.py
class ThisTester(ModelTester):
  def __init__(self, model_name, mode):
    self.model_name = model_name
    self.mode = mode
    self.model = xxx
    self.inputs = xxx

  def test_model_eval(self, as_ttnn=False, option=None):
    model = self.model
    inputs = self.inputs
    # code of set model & input as eval
    if as_ttnn == True:
        model = self.compile_model(model, option)
    # code of run model
    # code of get result
  def test_model_train(self, as_ttnn=False, option=None):
    ...
```
If make sure load model & inputs deterministic then its code can put in test_model_eval & test_model_train (but will load model & inputs again in compile_and_run)

swimdi:
Hi @kevinwuTT I think the current code can also do the thing you say
 - Can also pass record_property to avoid running fixture
 - use `test_model(self, as_ttnn=True, option=option)` will do the torch compile

kevinwuTT:
@swimdi

This will still require importing and modifying the test file around `ModelTester` right? My example was only relying on importing `torch_ttnn` which is what users who install the `torch_ttnn` package will do. `ModelTester` will be used for developers like us. I believe the `ttnn` package doesn't expose modules used for developing/testing like the `check_pcc` function for example. That means we can't use the test files outside of the testing environment. If we can re-use the model test files that will be nice. Otherwise, we will probably have to come up with some "demos" that are separate from tests. I was just wondering if you have any ideas if this is possible with `ModelTester`.

swimdi:
Hi @kevinwuTT,
Or I cancel the `ModelTester` class and flatten its inherit to all the test by using template?
For example, the `test_mnist.py` will be like this
```
import torch
import pytest
import torch_ttnn
from torchvision import transforms, datasets
from torch.utils.data import DataLoader

import torch.nn as nn
import torch.nn.functional as F

# adapted from https://github.com/pytorch/examples/blob/main/mnist/main.py
class MnistModel(torch.nn.Module):
    def __init__(self):
        super(MnistModel, self).__init__()
        self.conv1 = nn.Conv2d(1, 32, 3, 1)
        self.conv2 = nn.Conv2d(32, 64, 3, 1)
        self.dropout1 = nn.Dropout(0.25)
        self.dropout2 = nn.Dropout(0.5)
        self.fc1 = nn.Linear(9216, 128)
        self.fc2 = nn.Linear(128, 10)

    def forward(self, x):
        x = self.conv1(x)
        x = F.relu(x)
        x = self.conv2(x)
        x = F.relu(x)
        x = F.max_pool2d(x, 2)
        x = self.dropout1(x)
        x = torch.flatten(x, 1)
        x = self.fc1(x)
        x = F.relu(x)
        x = self.dropout2(x)
        x = self.fc2(x)
        output = F.log_softmax(x, dim=1)
        return output

class ThisTester:
    def __init__(self, model_name, mode):
        if mode not in ["train", "eval"]:
            raise ValueError(f"Current mode is not supported: {mode}")
        self.model_name = model_name
        self.mode = mode
        self.model = self._load_model()
        self.inputs = self._load_inputs()

    def _load_model(self):
        model = MnistModel()
        model = model.to(torch.bfloat16)
        return model

    def _load_inputs(self):
        transform = transforms.Compose([transforms.ToTensor()])
        test_dataset = datasets.MNIST(root="./data", train=False, transform=transform, download=True)
        dataloader = DataLoader(test_dataset, batch_size=1)
        test_input, _ = next(iter(dataloader))
        test_input = test_input.to(torch.bfloat16)
        return test_input


    def test_model_train(self, as_ttnn=False, option=None):
        # Fixing the random seed for reproducibility to ease debugging.
        #
        # Training processes involve more randomness compared to evaluation,
        # such as random initialization of weights.
        # Setting a fixed random seed is crucial for consistent testing
        # and debugging during the training process.
        torch.manual_seed(99)
        model = self.model
        model.train()
        model.zero_grad()

        # Eliminating randomness from Dropout to ensure consistent results.
        #
        # This is necessary for comparing the two training rounds:
        # one using PyTorch native code and the other using the PyTorch Dynamo TT backend.
        # Without this, the results would differ, making it impossible to compare the two implementations.
        for layer in model.modules():
            if isinstance(layer, torch.nn.Dropout):
                layer.p = 0  # Set dropout probability to 0

        inputs = self.inputs
        # Setting input tensor's `requires_grad` attribute to true.
        #
        # This allows us to use the gradient of the input as the golden result for the training process.
        # For further details, refer to the file `conftest.py` regarding the rationale behind.
        inputs.requires_grad_(True)

        if as_ttnn == True:
            model = torch.compile(model, backend=torch_ttnn.backend, options=option)
        outputs = model(inputs)
        loss = torch.mean(outputs)
        loss.backward()
        # Again, use the gradient of the input (`test_input.grad`) as the golden result for the training process.
        results = inputs.grad
        return results

    @torch.no_grad()
    def test_model_eval(self, as_ttnn=False, option=None):
        model = self.model.eval()
        inputs = self.inputs
        if as_ttnn == True:
            model = torch.compile(model, backend=torch_ttnn.backend, options=option)
        outputs = model(inputs)
        results = outputs
        return results

    def test_model(self, as_ttnn=False, option=None):
        if self.mode == "train":
            return self.test_model_train(as_ttnn, option)
        elif self.mode == "eval":
            return self.test_model_eval(as_ttnn, option)
        else:
            raise ValueError(f"Current mode is not supported: {self.mode}")


@pytest.mark.parametrize(
    "mode",
    ["train", "eval"],
)
def test_mnist_train(record_property, mode):
    model_name = "Mnist"
    record_property("model_name", f"{model_name} {mode}")

    tester = ThisTester(model_name, mode)
    results = tester.test_model()

    record_property("torch_ttnn", (tester, results))
```
And it will be generated by template like this

```
default_set_model_train = """
model.train()
model.zero_grad()

# Eliminating randomness from Dropout to ensure consistent results.
#
# This is necessary for comparing the two training rounds:
# one using PyTorch native code and the other using the PyTorch Dynamo TT backend.
# Without this, the results would differ, making it impossible to compare the two implementations.
for layer in model.modules():
    if isinstance(layer, torch.nn.Dropout):
        layer.p = 0  # Set dropout probability to 0
return model
"""

def export_model_test(file_path, load_model, load_input, set_model_train = default_set_model_train, ...):
    ...


mnist_load_model = """
model = MnistModel()
model = model.to(torch.bfloat16)
return model
"""

mnist_load_inputs = """
...
"""

export_model_test("test_mnist.py", load_model = mnist_load_model, load_input = mnist_load_inputs)

```

PR #276, opened 2024-10-01
Update precision for optimized trigonometric functions

jdh8:
### Ticket
Depends on tenstorrent/tt-llk-wh-b0#39

### Problem description
I have optimized implementation of trig functions.  It should result in better accuracy and probably run faster.

### What's changed
Test precision with stricter `assert_with_pcc`


ayerofieiev-tt (approved):

jdh8:
I made a stricter test case for `ttnn.tan` to reflect tenstorrent/tt-metal#14368, and it consistently passes locally.  Does CI uses the latest `tt-metal`?

PR #305, opened 2024-10-09
Remove xfail from Falcon model test

mcw-zwakeelTT:
### Problem description
Falcon model had many breakages, hence it was marked as _xfail_
### What's changed
Made the following fixes to enable model execution - 

- [ ] Model not fitting in DRAM (exceeding limit in layer norm)
- [ ] relational scalar ops (ttnn.eq/lt/le/gt/ge/ne)
- [ ] ttnn.full
- [ ] ttnn.arange
- [ ] ttnn.mul
- [ ] ttnn.arange op fixed in my forked tt-metal repo, will raise a PR

Added conversions for aten.eq.Scalar & aten.gt.Scalar, but for input variations specific to Falcon, its not converted (skipped due to errors).

### Constraints

- ttnn.full can't produce tensor shapes of kind [1,1], [1,1,1] etc throwing this error:
      - `Buffer size and page size should be larger than 0 bytes`
- ttnn.full can produce tilized tensors only if the shape is tilized, for non-tilized shapes, row major layout should be used.
- relational scalar ops and ttnn.mul require inputs to be tilized.
- inputs to elwise ops should be converted to tile layout before passing to their respective ops. 
- Input variation in ttnn.arange produced uint32 dtype tensor but the user of it was an elwise op (ttnn.mul) which required inputs to be tilized. `Constraint is that only bfloat16 dtype tensors can be tilized.` So for now, I have used bfloat16 dtype only for ttnn.arange rather than uint32


docs/models/Falcon/input_variations.md
```diff
@@ -1,199 +1,91 @@
...
-| 22 | aten.permute.default           |                  5 |           0 |         0 |          0 | ✘           |     0   |
-| 23 | aten.slice.Tensor              |                  8 |           0 |         0 |          0 | ✘           |     0   |
-| 24 | aten.t.default                 |                  1 |           0 |         0 |          0 | ✘           |     0   |
-| 25 | aten.transpose.int             |                  3 |           0 |         0 |          0 | ✘           |     0   |
-| 26 | aten.tril.default              |                  1 |           0 |         0 |          0 | ✘           |     0   |
-| 27 | aten.unsqueeze.default         |                  2 |           1 |         0 |          0 | 🚧          |     0.5 |
-| 28 | aten.view.default              |                 14 |           0 |         0 |          0 | ✘           |     0   |
-| 29 | aten.zeros_like.default        |                  1 |           0 |         0 |          0 | ✘           |     0   |
+|    | Operations              |   Input Variations |   Converted |   Removed |   Fallback | Completed   |   Score |
```
ayerofieiev-tt, line 2: There were 29 ops before, but now only 14 - why?

PR #308, opened 2024-10-11
Decompose `aten.masked_fill` to ttnn ops

kevinwuTT:
### Ticket
None

### Problem description
`aten.masked_fill` is used in Bloom. There does not seem to be a direct equivalent TTNN op.

### What's changed
`masked_fill` can be decomposed into a series of TTNN ops. However, this may be more unnecessarily complicated than simply generating a tensor on CPU and transferring the data to device. Nevertheless, the implementation is as follows:

If the signature of `aten.masked_fill` is `(Tensor tensor, Tensor mask, Scalar fill_value)` then the equivalent TTNN ops should be:

`ttnn_masked_fill = (tensor * (ttnn.ones - mask)) + (mask * ttnn.full(fill_value))`

- [x] Waiting on to merge https://github.com/tenstorrent/pytorch2.0_ttnn/pull/326 and update the docs with these changes


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -676,13 +676,47 @@ def rewrite_node(node):
...
                 target_users_ops = [user.target for user in node.users.keys()]
                 # Float and int types can be converted to ttnn.bfloat16, but bool may be problematic
+                # Can be removed if bool is only used for the following ops that have lowering:
+                ops_safe_to_remove = set([torch.ops.aten.masked_fill.Scalar])
+                can_remove_bool = kwargs["dtype"] != bool or (
+                    kwargs["dtype"] == torch.bool and ops_safe_to_remove.intersection(target_users_ops)
+                )
                 # Skip if type casting from bool and if the graph output uses this op
-                if kwargs["dtype"] not in [torch.bool] and "output" not in target_users_ops:
+                if can_remove_bool and "output" not in target_users_ops:
```
ayerofieiev-tt, line 685: Does not look like a part of masked_fill change

kevinwuTT: This is dependent on masked_fill, but can be separated into its own PR. I'll revert this part.

tests/lowering/creation/test_masked_fill.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def forward(self, input, mask, fill_value):
+        _to_copy = torch.ops.aten._to_copy.default(mask, dtype=torch.bool)
+        return input.masked_fill(_to_copy, fill_value)
+
+
+@pytest.mark.parametrize(
+    "input_shape, mask_shape, fill_value",
+    [
+        ((1, 1, 32, 32), (1, 1, 32, 32), -3.3895313892515355e38),
```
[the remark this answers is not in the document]
kevinwuTT: The range of float32 should be [-3.4e+38 to 3.4e+38](https://www.w3schools.com/go/go_float_data_type.php) so smallest negative value. For Bloom, I see this pattern repeated in the graph quite a bit.

```
masked_fill_14 = torch.ops.aten.masked_fill.Scalar(view_187, _to_copy_5, -3.3895313892515355e+38)                                           
_to_copy_30 = torch.ops.aten._to_copy.default(masked_fill_14, dtype = torch.float32, layout = torch.strided, device = device(type='cpu'))
_softmax_12 = torch.ops.aten._softmax.default(_to_copy_30, -1, False)
```

ayerofieiev-tt (approved):

PR #323, opened 2024-10-18
Unmark compilation_xfail after generated guard function updated

swimdi:
### Ticket
N/A

### Problem description
Before the guard function update, the statistic of model test is:
 - total model tests: 152
 - passed model tests: 27
 - failed model tests: 125

[After the guard function update](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/321), the statistic of model test is:
 - total model tests: 152
 - passed model tests: 45
 - failed model tests: 107

So umark compilation_xfail of newly pass case

### What's changed
unmark compilation_xfail of these models:
```
Autoencoder (linear) eval
Autoencoder (linear) train
dla34.in1k train
facebook/deit-base-patch16-224 eval
HardNet train
OpenPose V2 train
Perceiver IO eval
ResNet18 train
ResNet50 train
Unet-brain train
Unet-carvana train
U-Net train
vit_b_16 eval
vit_b_32 eval
vit_h_14 eval
vit_l_16 eval
vit_l_32 eval
xception71.tf_in1k train
```

### DIFF
--- a/tests/models/timm/test_timm_image_classification.py
+++ b/tests/models/timm/test_timm_image_classification.py
@@ -34,32 +34,45 @@ class ThisTester(ModelTester):
         return input_batch
 
 
-@pytest.mark.parametrize(
-    "mode",
-    ["train", "eval"],
-)
+model_list = [
+    "tf_efficientnet_lite0.in1k",
+    "tf_efficientnet_lite1.in1k",
+    "tf_efficientnet_lite2.in1k",
+    "tf_efficientnet_lite3.in1k",
+    "tf_efficientnet_lite4.in1k",
+    "ghostnet_100.in1k",
+    "ghostnetv2_100.in1k",
+    "inception_v4.tf_in1k",
+    "mixer_b16_224.goog_in21k",
+    "mobilenetv1_100.ra4_e3600_r224_in1k",
+    "ese_vovnet19b_dw.ra_in1k",
+    "xception71.tf_in1k",
+    "dla34.in1k",
+    "hrnet_w18.ms_aug_in1k",
+]
+mode_list = ["train", "eval"]
+model_and_mode_list = []
+for model in model_list:
+    for mode in mode_list:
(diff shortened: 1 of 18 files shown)


tests/models/timm/test_timm_image_classification.py
```diff
@@ -34,32 +34,45 @@ def _load_inputs(self):
...
+    "inception_v4.tf_in1k",
+    "mixer_b16_224.goog_in21k",
+    "mobilenetv1_100.ra4_e3600_r224_in1k",
+    "ese_vovnet19b_dw.ra_in1k",
+    "xception71.tf_in1k",
+    "dla34.in1k",
+    "hrnet_w18.ms_aug_in1k",
+]
+mode_list = ["train", "eval"]
+model_and_mode_list = []
```
ayerofieiev-tt, line 54: It is fairly hard to understand what’s marked with fail from this. 
sometimes it is better to write plain dumb code - let’s do this here so it’s very clear. Wdyt?

ayerofieiev-tt (approved):

ayerofieiev-tt:
What are the next steps to enable the rest of models?

swimdi:
Because there still many failed model test, and I think currently it is difficult to increase the pass rate, so the next step I want to do is to debug the failed input variation case

Apparently, there have two type of failed case
 - failed in single op test: this is already shown in guard function, can debug directly
 - pass in single op test, but failed in model test: I think I can find them out by doing these steps
    1. disable all lowerings to ensure all model pass
    2. enable one lowering and check if any model failed, and start debugging the failed input variation until it is fixed or updated to guard function, and then keep it on and move to next lowering

debug first type of failed case theoretically not increase the pass rate.

debug or guard second type seems can increase the pass rate, but becuase the method I do is start by the zero lowering situation, so I think it cannot see the effect in a short time.

which type of failed case you prefer to do first @@?

PR #326, opened 2024-10-19
[CI] Fix failed of collecting metrics

swimdi:
### Ticket
N/A

### Problem description
 - [CI failed to collect metrics](https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/11411386896/job/31758906781)
 - Want to remove eval from models

### What's changed
 - Use `merge-multiple: true` to correct folder structure of `metrics-input-variations`
 - Change model test name `xxx eval` back to `xxx`

### DIFF
--- a/.github/workflows/before_merge.yaml
+++ b/.github/workflows/before_merge.yaml
@@ -94,7 +94,7 @@ jobs:
           merge-multiple: true
           path: metrics/
 
-  model-input-variations-tests:
+  model-autogen-op-tests:
     needs: model-tests
     runs-on: ["in-service", "n150"]
     strategy:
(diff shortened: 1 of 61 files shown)


ayerofieiev-tt (commented):
Let’s move this if to the infra

    if mode == "eval":
        record_property("model_name", model_name)
    else:
        record_property("model_name", f"{model_name} {mode}")


Instead in models we can pass 
record_property("model_name",  mode)

and infra can compose a proper name. This will drastically reduce boilerplate across models

.github/workflows/before_merge.yaml
```diff
@@ -115,6 +116,7 @@ jobs:
       - name: Run Model Input Variations Tests
         run: |
           set +e
+          rm -rf tests/input-variations/
```
ayerofieiev-tt, line 119: Let’s indicate that this folder is autogenerated so nobody changes files there. 
Something like
tests/autogen-op-tests/

swimdi:
I haved move the model name decision in `confest.py`

ayerofieiev-tt (approved):

PR #332, opened 2024-10-23
Leverage CI running time of autogen-op test 

swimdi:
### Ticket
N/A

### Problem description
The `tests/autogen-op/Stable_Diffusion_V2` tests is running in same CI job, but `Stable_Diffusion_V2` spent many time, want to leverage it

### What's changed
 - Original method of autogen-op is split CI job model by model, now change to split op by op, so tests in `tests/autogen-op/Stable_Diffusion_V2` will be dispatch to all 40 job, not only run at same CI job
 - Fix and add feature in `tools/generate_input_variation_test_from_models.py`
### Question
The test scope of autogen-op basically same with model test and it just used in `docs/`, so or I just let autogen-op enable only if `${{ github.event_name == 'workflow_dispatch' && github.event.inputs.commit_report != 'None'}}` ?

### DIFF
--- a/.github/workflows/before_merge.yaml
+++ b/.github/workflows/before_merge.yaml
@@ -91,7 +91,6 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: model-tests-metrics-group-${{ matrix.group }}
-          merge-multiple: true
           path: metrics/
 
   model-autogen-op-tests:
@@ -108,10 +107,11 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/common_repo_setup
 
-      - name: Download Metrics Artifacts for Generate Input Variations Tests
+      - name: Download All Metrics Artifacts for Generate Input Variations Tests
         uses: actions/download-artifact@v4
         with:
-          pattern: model-tests-metrics-group-${{ matrix.group }}
+          pattern: model-tests-metrics-group-*
+          merge-multiple: true
           path: metrics/
       - name: Run Model Input Variations Tests
         run: |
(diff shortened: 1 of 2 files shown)


ayerofieiev-tt (commented):
Maybe we must split creation of these tests into a separate thing?
If these tests created/updated separately from time to time and stored in the repo, we could then start running them in parallel.

swimdi:
Ok, I can manually update `tests/autogen_op` by PR, then the single op test and model test can run on the same stage.

Beside, this PR seems not reduce the running time...
 - [Original running method](https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/11461760822) has one job spent 42 min (model-autogen-op-tests 16)
 - [This PR's running method](https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/11475386829) have two job spent 48m & 37m (model-autogen-op-tests 10 & 19)
 
So I think I will cancel this PR becasue the new running method is worse
 
And I found the original running method, the CI job of spent 42 min, mostly spent from `tests/autogen-op/Stable_Diffusion_V2/test_Stable_Diffusion_V2_aten_convolution_default.py` (I run it and it spent 38 min)

So here is my new propose
1. will open a PR to add `tests/autogen_op/`
2. let `model-tests` & `model-autogen-op-tests` parallel run in the same stage
3. skip running `tests/autogen-op/Stable_Diffusion_V2/test_Stable_Diffusion_V2_aten_convolution_default.py` so the job can reduce time spent from 42 min to few min
4. If don't want to run `model-autogen-op-tests` every time, I can let it triggered only if `${{ github.event_name == 'workflow_dispatch' && github.event.inputs.commit_report != 'None'}}`

PR #337, opened 2024-10-24
Add autogen op tests

swimdi:
### Ticket
N/A

### Problem description
As mentioned in [here](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/332#issuecomment-2434700680), add autogen op tests

### What's changed
 - Add `tests/autogen_op`
 - Skip running tests/autogen-op/Stable_Diffusion_V2/test_Stable_Diffusion_V2_aten_convolution_default.py so the job can reduce time spent from 42 min to few min


ayerofieiev-tt:
I am concerned how adding 2k small files affects repo performance. But I am ok to merge this now as is. 

Please think if it would be better to store raw collected data and generate tests on the fly from that.

ayerofieiev-tt (approved):
Later instead of storing generated tests per model, we can store raw data and generate tests on the fly. I am a bit concerned how about adding 2k

swimdi:
`tests/autogen-op` is generated by `metrics/` folder, which is the record generated from model tests.

If store the `metrics/`(i.e. the raw data) in the fly, then can `python3 tools/generate_input_variation_test_from_models.py` to generate `tests/autogen-op`

I don't know how to add either raw data or `tests/autogen-op` on the fly @@

PR #338, opened 2024-10-24
CI parallel autogen op

swimdi:
### Ticket
N/A

### Problem description
As mentioned in [here](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/332#issuecomment-2434700680), let `model-tests` & `model-autogen-op-tests` parallel run in the same stage

### What's changed
 - Let `model-tests` & `model-autogen-op-tests` parallel run in the same stage
 - Let `model-autogen-op-tests` triggered only if ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.commit_report != 'None'}}

### DIFF
--- a/.github/workflows/before_merge.yaml
+++ b/.github/workflows/before_merge.yaml
@@ -94,7 +94,7 @@ jobs:
           path: metrics/
 
   model-autogen-op-tests:
-    needs: model-tests
+    needs: lowering-tests
     runs-on: ["in-service"]
     strategy:
       matrix: # Need to find a way to replace this with a generator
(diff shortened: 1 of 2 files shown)


ayerofieiev-tt (approved):

.github/workflows/before_merge.yaml
```diff
@@ -108,21 +108,10 @@ jobs:
...
         run: |
           set +e
-          rm -rf tests/autogen-op/
-          python3 tools/generate_input_variation_test_from_models.py
-          exit_code=$?  # Capture the exit code
-          if [ $exit_code -ne 0 ]; then
-            echo "Failed to generate input variation test : exit code $exit_code.";
-            exit 1;  # Fail the workflow for other errors
-          fi
+          rm -rf tests/autogen-op/ALL
```
ayerofieiev-tt, line 114: Why?

swimdi: The `ALL` is the merged of all models, that is, if aten.add.Tensor have input_var_A in modelA and inuput_var_B in modelB, then the aten.add.Tensor test in `ALL` will both contain input_var_A and input_var_B

`python3 -m pytest --github-report tests/autogen-op -s` will run all the test include `ALL` and models, so I remove `tests/autogen-op/ALL` is for avoid running duplicate case.

PR #369, opened 2024-10-31
Add Qwen2.5, Gemma-1, Phi-2

cglagovichTT:
I'm adding support for a few more language models which are high priority for the models team. 

Note that the Gemma model requires a huggingface-cli login. In addition, Qwen2-VL is marked to be skipped since the transformers version is < 4.45. 

@kevinwuTT could you help me understand why the Qwen and Gemma tests fail like this? It doesn't look like these tests are any different from Phi-2 and GPT-Neo, which pass.
```
E                   TypeError: Gemma-1.1-2b-it - Torch run with bypass compilation failed. Please check whether `model` or `model.generate` is passed to `record_property`.
```

### DIFF
--- a/tests/models/gemma_2b_it/test_gemma_2b_it.py
+++ b/tests/models/gemma_2b_it/test_gemma_2b_it.py
@@ -1,0 +1,44 @@
+import torch
+import pytest
+from transformers import AutoTokenizer, AutoModelForCausalLM
+from tests.utils import ModelTester
+
+
+class ThisTester(ModelTester):
+    def _load_model(self):
+        model_name = "google/gemma-1.1-2b-it"
+        self.tokenizer = AutoTokenizer.from_pretrained(
+            model_name, padding_side="left", torch_dtype=torch.bfloat16
+        )
+        model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.bfloat16)
+        return model.generate
+
+    def _load_inputs(self):
+        prompt = "This is a test prompt."
+        input_ids = self.tokenizer(prompt, return_tensors="pt").input_ids
+        arguments = {
+            "input_ids": input_ids,
+            "do_sample": True,
+            "temperature": 0.9,
+            "max_new_tokens": 5,
+        }
+        return arguments
+
+    def set_model_eval(self, model):
+        return model
+
+
+@pytest.mark.parametrize("mode", ["eval"])
+@pytest.mark.compilation_xfail
+def test_gemma(record_property, mode):
+    model_name = "Gemma-1.1-2b-it"
+    record_property("model_name", model_name)
+    record_property("mode", mode)
+
+    tester = ThisTester(model_name, mode)
+    results = tester.test_model()
+    if mode == "eval":
+        gen_text = tester.tokenizer.batch_decode(results)[0]
+        print(f"Generated Text:\n{gen_text}")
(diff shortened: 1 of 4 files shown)


cglagovichTT:
fyi @uaydonat

npetrovic-tenstorrent:
> fyi @uaydonat

For Qwen2 test, it returns  many warnings of this type. I also attach full log. To reproduce:
pytest [tests/models/qwen/test_qwen.py  -s &> /tmp/qwen2.log
[log_qwen2.log](https://github.com/user-attachments/files/19328478/log_qwen2.log)

 [2025-03-18 21:20:13,936] torch._dynamo.convert_frame: [WARNING] torch._dynamo.exc.BackendCompilerFailed: backend=‘ttnn_backend’ raised:
[2025-03-18 21:20:13,936] torch._dynamo.convert_frame: [WARNING] RuntimeError: aten::clone() Expected a value of type ‘Tensor’ for argument ‘self’ but instead found type ‘SymInt’.
[2025-03-18 21:20:13,936] torch._dynamo.convert_frame: [WARNING] Position: 0
[2025-03-18 21:20:13,936] torch._dynamo.convert_frame: [WARNING] Value: s0
[2025-03-18 21:20:13,936] torch._dynamo.convert_frame: [WARNING] Declaration: aten::clone(Tensor self, *, MemoryFormat? memory_format=None) -> Tensor
[2025-03-18 21:20:13,936] torch._dynamo.convert_frame: [WARNING] Cast error details: Unable to cast s0 to Tensor
[2025-03-18 21:20:13,936] torch._dynamo.convert_frame: [WARNING]

npetrovic-tenstorrent:
> n with bypass compilation failed. Please check whether `model` or `model.generate` is passed to `record_proper

After more detailed investigation -  it seems that for both Qwen2 and Qwen2.5 there is some problem with torch.compile itself, as even in case when TTNN backend is not used for compilation, the same error occurs. Looks like  torch.compile does not behave properly with either the qwen model or newer versions of transformers. transformers>=4.45.0 .

kevinwuTT:
For this particular error concerning `is_compiling`, there is a difference in implementation between v4.38.0 and v4.45.0:

v4.38.0
https://github.com/huggingface/transformers/blob/08ab54ada594f8f4cc1e458b1217bf8c53f04dd8/src/transformers/utils/import_utils.py#L529-L537

```
def is_torchdynamo_compiling():
    if not is_torch_available():
        return False
    try:
        import torch._dynamo as dynamo  # noqa: F401

        return dynamo.is_compiling()
    except Exception:
        return False
```

v4.45.0
https://github.com/huggingface/transformers/blob/2ef31dec1676249d26044a8aa8abe33dbecf0d10/src/transformers/utils/import_utils.py#L734-L750

```
def is_torchdynamo_compiling():
    if not is_torch_available():
        return False

    # Importing torch._dynamo causes issues with PyTorch profiler (https://github.com/pytorch/pytorch/issues/130622)
    # hence rather relying on `torch.compiler.is_compiling()` when possible (torch>=2.3)
    try:
        import torch

        return torch.compiler.is_compiling()
    except Exception:
        try:
            import torch._dynamo as dynamo  # noqa: F401

            return dynamo.is_compiling()
        except Exception:
            return False
```

You can try to investigate why the newer implementation is causing problems with `torch.compile`.

PR #379, opened 2024-11-05
Implement ConstantFoldingPass

pschang-phy:
### Ticket
N/A

### Problem description

some models contain operations that have known constant arguments (inputs) before compiling, such operations could be transformed (fold) into constant values to simplify model graph.

### What's changed

Implement a `ConstantFolder` that handles folded operations.

For now, support operations are
- torch.ops.aten.lift_fresh_copy.default,
- torch.ops.aten.pow.Tensor_Tensor
- torch.ops.aten.arange.start
- torch.ops.aten.unsqueeze.default

cc @swimdi @jdh8

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -14,6 +14,7 @@ import numpy as np
 from typing import Tuple
 import torch_ttnn.metrics as metrics
 
+from torch._subclasses.fake_tensor import unset_fake_temporarily
 from torch.fx.passes.infra.pass_base import PassBase, PassResult
 import torch.fx.traceback as fx_traceback
 from . import target_wrappers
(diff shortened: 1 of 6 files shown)


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -768,12 +769,77 @@ def rewrite_node(node):
     return gm
 
 
+class ConstantFolder(torch.nn.Module):
```
ayerofieiev-tt, line 772: Can we move to a separate file? This one is growing way out of comfort zone

boris-drazic: Yeah, I think constant folding should be a separate compiler pass and in `backend.py` called before `ToTtPass`.

pschang-phy: I have moved to separate `ConstantFoldingPass` and apply before `ToTtPass`. refer to 4d3af2bead5d1da95e0ed3d4f601f00c62912024

ayerofieiev-tt (approved):

Issue #377, opened 2024-11-04
`ttnn.permute` fails with inner-most dim = 1 when doing transpose


jerrysky3:
`ttnn.permute` fails with inner-most dim = 1 when doing transpose (tested with 2D tensors, e.g. `(2, 1)` with permutation `(1, 0)`).

This is blocking the `aten.t.default` to `ttnn.permute` conversion in models:

- `CLIP`
- `CLIP-train`

**To reproduce**

Run the example below:
```py
import torch
import ttnn


def main(device):
    torch_tensor = torch.rand((2, 1), dtype=torch.bfloat16)
    print(torch_tensor)
    input_tensor = ttnn.from_torch(torch_tensor, layout=ttnn.TILE_LAYOUT, device=device)
    input_tensor = ttnn.permute(input_tensor, (1, 0))
    print(ttnn.to_torch(input_tensor), input_tensor.shape)


if __name__ == "__main__":
    device = ttnn.open_device(device_id=0)
    try:
        main(device)
    finally:
        ttnn.close_device(device)
```

Currently it will fail with the error:

```
RuntimeError: TT_FATAL @ ../ttnn/cpp/ttnn/tensor/types.cpp:211: normalized_index >= 0 and normalized_index < rank
info:
Index is out of bounds for the rank, should be between 0 and 0 however is 18446744073709551615
backtrace:
 --- /home/jerry/tt-metal/ttnn/ttnn/_ttnn.so(+0x100ef49) [0x7f53b10def49]
 --- ttnn::types::Shape::operator[](long) const
 --- ttnn::operations::data_movement::ReshapeViewOperation::invoke(tt::tt_metal::Tensor const&, ttnn::types::Shape const&)
 --- /home/jerry/tt-metal/ttnn/ttnn/_ttnn.so(+0x4e7f92) [0x7f53b05b7f92]
 --- ttnn::operations::data_movement::ExecutePermute::invoke(unsigned char, tt::tt_metal::Tensor const&, tt::stl::Span<long const, 18446744073709551615ul>, std::__1::optional<tt::tt_metal::MemoryConfig> const&, bool)
 --- /home/jerry/tt-metal/ttnn/ttnn/_ttnn.so(+0x12a60cb) [0x7f53b13760cb]
 --- /home/jerry/tt-metal/ttnn/ttnn/_ttnn.so(+0x11fad8b) [0x7f53b12cad8b]
[...]

PR #381, opened 2024-11-06
Enable `aten.t.default` conversion for more cases

jerrysky3:
### Ticket
#377

### Problem description
Enable `aten.t.default` conversion for more cases and fallback for the issue mentioned in #377

### What's changed
- Explicitly handle unsupported cases in `to_tt_pass.py`
- Remove `aten.t.default` from blocklist
- Add tests to verify support and unsupported cases

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -634,16 +634,27 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool
                     return g.call_function(ttnn.reshape, args=(args[0], output_size))
                 return None
 
-            if node.target == torch.ops.aten.transpose.int:
-                dim0 = args[1]
-                dim1 = args[2]
+            if node.target in [torch.ops.aten.transpose.int, torch.ops.aten.t.default]:
                 output_size = node.meta["val"].size()
                 rank = len(output_size)
+                if node.target == torch.ops.aten.t.default:
+                    assert rank >= 0 and rank <= 2, "Input tensor can only be 0D, 1D or 2D"
+                    if rank < 2:
+                        # Less 2D transpose is no-op
+                        return args[0]
+                    dim0 = 0
+                    dim1 = 1
+                else:
+                    dim0 = args[1]
+                    dim1 = args[2]
                 permutation = list(range(rank))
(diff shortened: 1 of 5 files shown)


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -634,16 +634,27 @@ def rewrite_node(node):
                     return g.call_function(ttnn.reshape, args=(args[0], output_size))
                 return None
 
-            if node.target == torch.ops.aten.transpose.int:
-                dim0 = args[1]
-                dim1 = args[2]
+            if node.target in [torch.ops.aten.transpose.int, torch.ops.aten.t.default]:
                 output_size = node.meta["val"].size()
                 rank = len(output_size)
+                if node.target == torch.ops.aten.t.default:
```
ayerofieiev-tt, line 640: what's the difference between `torch.ops.aten.transpose` and `ops.aten.t`?

jerrysky3: I guess there are historical reasons but `aten.t` only transposes 2D tensor and `aten.transpose` can swap any two dims on n-D tensor

https://pytorch.org/docs/stable/generated/torch.t.html
https://pytorch.org/docs/stable/generated/torch.transpose.html

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -634,16 +634,27 @@ def rewrite_node(node):
...
+                    assert rank >= 0 and rank <= 2, "Input tensor can only be 0D, 1D or 2D"
```
ayerofieiev-tt, line 641: instead of asserting, should we fallback?

jerrysky3: Given the pytorch spec, this shouldn't be possible: https://pytorch.org/docs/stable/generated/torch.t.html

"Expects input to be <= 2-D tensor and transposes dimensions 0 and 1."

So I think it makes sense to use assert here

ayerofieiev-tt: Thank you!

ayerofieiev-tt (approved):

torch_ttnn/passes/lowering/to_tt_pass.py
ayerofieiev-tt: noted, thank you

PR #383, opened 2024-11-06
Ensure to_torch with correct dtype

swimdi:
### Ticket
#360 (https://github.com/tenstorrent/tt-metal/issues/14493)

### Problem description
`try_add_data_move_in` will change the dtype to `TtnnBfloat16`(or `TtnnUint32` if user is emb), which break the dtype consistency. In ttnn I think it is fine because ttnn usually accept `TtnnBfloat16` dtype, but in aten torch, some are not

Take `aten.index_select.default` for example:
```
arange = torch.ops.aten.arange.start(0, 19, dtype=torch.int64, device=torch.device(type="cpu"), pin_memory=False)
add = torch.ops.aten.add.Tensor(arange, 2)
clamp = torch.ops.aten.clamp.default(unsqueeze, 0)
index_select = torch.ops.aten.index_select.default(x, 0, clamp)
```
The third input of `index_select`, i.e. tensor `clamp`, only accept integer dtype, but when this code convert to ttnn, the `torch.ops.aten.add.Tensor` is replaced as `ttnn.add` and its output become `TtnnBfloat16` due to `try_add_data_move_in`, and `clamp` dtype become `TtnnBfloat16`

Although `AddDataMovePass::ensures` try to solve this issue, but if `clamp` is not converted to ttnn, then because `AddDataMovePass::ensures` only detect `node.target == ttnn.to_torch`, so it cannot handle this issue

So the fix is to add `aten._to_copy.default` at `try_add_data_move_out*` with correct dtype from `node.meta["val"]`, so it can ensure the torch tensor type is correct (if `to_torch` can specifiy dtype, then can merge `to_copy` on it)

Because `aten._to_copy.default` only added for the fallback aten op case, so it won't effected if model is end to end converted

### What's changed
1. Remove `AddDataMovePass::ensures`
2. Add `aten._to_copy.default` at `try_add_data_move_out*` to replace the role of `AddDataMovePass::ensures`
3. Add condition of keeping aten._to_copy in to_tt_pass
4. Remove the blocklist related to issue #360


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -720,14 +723,24 @@ def rewrite_node(node):
...
                     return None
+                src_dtype = node.args[0].meta["val"].dtype
+                dst_dtype = kwargs["dtype"]
+                # Some aten op need it to cast specific dtype (ex, index_select)
+                # Keep it if casting from int to float or reverse
+                if dst_dtype in [torch.int32, torch.int64] and src_dtype not in [torch.int32, torch.int64]:
+                    return None
+                if src_dtype in [torch.int32, torch.int64] and dst_dtype not in [torch.int32, torch.int64]:
+                    return None
+                target_users_ops = [user.target for user in node.users.keys()]
```
kevinwuTT, line 741: This line probably not needed?

kevinwuTT (approved):

ayerofieiev-tt (approved):

torch_ttnn/passes/lowering/to_tt_pass.py
swimdi: I had removed it.

I wanted to avoid adding `aten._to_copy` to the user's output at the beginning. Because including it would mean that the model would always have at least one `aten._to_copy` operation, so that cannot achieve end to end compilation.

However, Falcon uses multiple graphs, and in some cases, one graph's output serves as the third index selection for another graph. In these cases, `aten._to_copy` is still needed to preserve the dtype.

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -378,9 +389,25 @@ def try_add_data_move_out_for_list_args(src_nodes, dst_idx, dst_node):
...
+                # TODO: After to_torch has dtype args, then can adjust back in to_torch
+                target_users_ops = [user.target for user in src_node.users.keys()]
+                if (
+                    "output" not in target_users_ops
+                    and hasattr(src_node, "meta")
+                    and "val" in src_node.meta
+                    and hasattr(src_node.meta["val"], "dtype")
+                ):
+                    src_meta_dtype = src_node.meta["val"].dtype
+                    node_copy = call_aten__to_copy_with_meta(g, node_to_torch, src_meta_dtype)
```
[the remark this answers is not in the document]
swimdi: ok, I had condense it to try_call_aten__to_copy_with_meta

PR #389, opened 2024-11-07
Lowering `torch.ops.aten.expand.default` to `ttnn.expand`

pschang-phy:
### Ticket
N/A

### Problem description
Some `torch.ops.aten.expand.default` ops exist.

### What's changed
If the `torch.ops.aten.expand.default` op has same number of elements in  input and output tensors, we could just lowering to `ttnn.reshape`.


torch_ttnn/passes/constant_folding_pass.py
```diff
@@ -0,0 +1,63 @@
+import torch
+from torch._subclasses.fake_tensor import unset_fake_temporarily
+from torch.fx.passes.infra.pass_base import PassBase, PassResult
+
+
+class ConstantFoldingPass(PassBase):
```
ayerofieiev-tt, line 6: This is a part of the other change. Is that expected?

torch_ttnn/passes/lowering/to_tt_guard.py
```diff
@@ -480,7 +480,7 @@
 # add output is bf16 type, which is wrong
 
 aten_slice_Tensor_blocklist += [
-    ["Tensor<[24]> self = ?", "int dim = 0", "Optional[int] start = 0", "Optional[int] end = 9223372036854775807"],
```
ayerofieiev-tt, line 483: why is this changed?

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -560,10 +560,17 @@ def rewrite_node(node):
                 return None
 
             if node.target == torch.ops.aten.expand.default:
+                input_tensor_shape = args[0].meta["val"].size()
```
ayerofieiev-tt, line 563: Are you aware about `ttnn.expand`?

ayerofieiev-tt: https://github.com/tenstorrent/tt-metal/blob/main/tests/ttnn/unit_tests/test_expand.py

ayerofieiev-tt: Keep in mind, I don't think its possible to convert this at 0 cost for tiles with tile padding. And because we don't have a support for a strided view - it is currently not possible at 0 cost even for Row Major tensors. 

Please try lowering to ttnn.expand.

torch_ttnn/passes/constant_folding_pass.py
pschang-phy: Because this branch is base on [dev/constant_folding](https://github.com/tenstorrent/pytorch2.0_ttnn/tree/dev/constant_folding),  the changes also show here.

I have switched the base of this branch.
Thank you for pointing that out.

torch_ttnn/passes/lowering/to_tt_guard.py
pschang-phy: Same as this https://github.com/tenstorrent/pytorch2.0_ttnn/pull/389#discussion_r1833651332

torch_ttnn/passes/lowering/to_tt_pass.py
pschang-phy: Thank you for remind about `ttnn.expand`.

### Test `ttnn.expand`
I try the `ttnn.expand` using https://github.com/tenstorrent/tt-metal/blob/main/tests/ttnn/unit_tests/test_expand.py, and notice that it fails to **torch.bfloat16** on certain cases due to page alignment. as below
```
function_args = (ttnn.Tensor([[ 0.53125,  0.97656,  ...,  0.00000,  0.00000],
             [ 0.00000,  0.00000,  ...,  0.00000,  0.000...0000,  ...,  0.00000,  0.00000]], shape=Shape([1[32], 3[32]]), dtype=DataType::BFLOAT16, layout=Layout::TILE), [3, -1])
function_kwargs = {}

    def __call__(self, *function_args, **function_kwargs):
>       return self.function(*function_args, **function_kwargs)
E       RuntimeError: TT_THROW @ /tmp/build-via-sdist-euf95zo8/metal_libs-0.53.0rc35+wormhole.b0/tt_metal/impl/buffers/circular_buffer_types.hpp:71: tt::exception
E       info:
E       Page size must be divisible by sizeof(uint32_t) because buffers holds uint32_t values
E       backtrace:
```

### Lowering to `ttnn.expand` in certain condition
Refer to 7aa9c4090284611cb543c31668f6f1681c8f342d, I have fix to use `ttnn.expand` with certain condition.

`ttnn.expand` can be used in the condition of **input_shape[0] %2 == 0**, test as below
```bash
$ pytest -s tests/lowering/misc/test_expand.py
```

ayerofieiev-tt (approved):

Issue #390, opened 2024-11-07
ViLT failed because `aten.add` & `ttnn.add` shape mismatch


swimdi:
# Brief describe
add shape mismatch for this input variation
```
"Tensor<[5]> self = ?", "Tensor other = 0.0"
```

Run [`pytest tests/pattern/test_add_shape_mismatch.py`](https://github.com/tenstorrent/pytorch2.0_ttnn/blob/stage1_swimdi_add/tests/pattern/test_add_shape_mismatch.py) can reproduce
```
 assert torch.Size([5]) == torch.Size([1, 5])
```

# Debug log
In ViLT, there have the following code
```
# /home/ubuntu/venv_pt/lib/python3.8/site-packages/transformers/models/vilt/modeling_vilt.py(122)<listcomp>()
    def visual_embed(self, pixel_values, pixel_mask, max_image_length=200):
        _, _, ph, pw = self.patch_embeddings.projection.weight.shape

        x = self.patch_embeddings(pixel_values)
        x_mask = pixel_mask[:, None, :, :].float()
        x_mask = nn.functional.interpolate(x_mask, size=(x.shape[2], x.shape[3])).long()
        x_h = x_mask[:, 0].sum(dim=1)[:, 0]
        x_w = x_mask[:, 0].sum(dim=2)[:, 0]

        batch_size, num_channels, height, width = x.shape
        patch_dim = self.config.image_size // self.config.patch_size
        spatial_pos = self.position_embeddings[:, 1:, :].transpose(1, 2).view(1, num_channels, patch_dim, patch_dim)
        pos_embed = torch.cat(
            [
                nn.functional.pad(
                    nn.functional.interpolate(
                        spatial_pos,
                        size=(h, w),
                        mode="bilinear",
                        align_corners=True,
                    ),
                    (0, width - w, 0, height - h),
                )
                for h, w in zip(x_h, x_w)
            ],
            dim=0,
        )
```

The original value of `x_h` and `x_w` are
```
x_h: TorchTensor([12])
x_w: TorchTensor([16])
```
But after convert to ttnn, the wrong value become
```
x_h: TorchTensor([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]])
x_w: TorchTensor([[12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12]])
```

[...]

Issue #358, opened 2024-10-30
ttnn.add failed with Inputs[1].shape = [1, 12[32], 768]


swimdi:
When I debug `albert-base-v2-classification` model test, has this error message
```
self = FastOperation(python_fully_qualified_name='ttnn.add', function=<ttnn._ttnn.operations.binary.add_t object at 0x7f0a5c8...<function default_postprocess_golden_function_outputs at 0x7f0a5c35bd30>, is_cpp_operation=True, is_experimental=False)
function_args = (ttnn.Tensor([[[ 0.74219,  0.07324,  ...,  1.32812,  0.25391],
              [ 0.57031,  0.38086,  ...,  0.37695, -0.1..., ttnn.Tensor(<buffer is not allocated>, shape=Shape([1, 12[32], 768]), dtype=DataType::BFLOAT16, layout=Layout::TILE))
function_kwargs = {}

    def __call__(self, *function_args, **function_kwargs):
>       return self.function(*function_args, **function_kwargs)
E       RuntimeError: TT_THROW @ /tmp/build-via-sdist-c9nw8bov/metal_libs-0.53.0rc16+wormhole.b0/ttnn/cpp/ttnn/tensor/tensor.hpp:263: tt::exception
```
It failed at `ttnn.add` and its input shape is
 - Inputs[0].shape = [1, 12, 768]
 - Inputs[1].shape = [1, 12[32], 768]

I'm not sure why the second shape has `12[32]` value

And if I block the lowering of this `ttnn.add` op back to `aten.add.Tensor`, then this test pass

The reproduce step is
1. Remove this line in `to_tt_guard.py`
```
aten_add_Tensor_blocklist += [["Tensor<[1, 12, 768]> self = ?", "Tensor<[1, 12, 768]> other = ?"]]
```
2.  `pytest tests/models/albert/test_albert_token_classification.py`

After this issue is resolved, please also remove the related blocklist in `to_tt_guard.py`.

swimdi:
I'm now trying to write a simpler pattern test that can reproduce this error

ayerofieiev-tt:
The issue here looks like its related to "ttnn.Tensor(<buffer is not allocated>"

jerrysky3:
It might due to the first operand of `ttnn.add` is a host tensor instead of the device tensor (while the second operand is a device tensor)

I found the model IR contains the following chain:

```py
[...]

PR #392, opened 2024-11-07
Resolved layout mismatch problem

swimdi:
### Ticket
#358 

### Problem description

When src_node is aten and dst_node is ttnn, it needs a `from_torch` node between, and if there are multiple ttnn dst_node use same aten src_node, current implement will reuse the `from_torch` node. That is
 - If src_node is aten and dst_node is ttnn, then will 
   - create a from_torch node and record in hash table
   - or get a from_torch node from hash table

But this implement has bug. For example, if `ttnn.add` and `ttnn.emb` use the same src_node, then they both needs `from_torch` node but with `"different layout"`. Current implement just give then `from_torch` node but with `"same layout"`. The layout type is decided by which ttnn.add or ttnn.emb is traversed first.

@jerrysky3 solved this problem by separating it into two stages
1. If src_node is aten and dst_node is ttnn, then will 
   - create a from_torch node with `"default layout"` and record in hash table
   - or get a from_torch node from hash table
2. If dst_node's layout requirement is NOT the `"default layout"`, then add an extra layout node to satisfy it

### What's changed
1.  cherry-pick [1c04d79](https://github.com/tenstorrent/pytorch2.0_ttnn/commit/1c04d79e01c0b54fe020a73a43d254e0ffba7422#diff-84bced1b9706f48f1e5ec14b61fed648f086f1ea6d5d67cd8243044a8c8e6d7c) to solved layout issue
2. Remove the blocklist related to issue #358 
3. Move `aten.add.Tensor` restricted from `to_tt_guard_autogen` to `to_tt_pass`
4. Add tests for #390

I develop this branch from #383, so I'll merge #383 first and then this PR


tests/conftest.py
```diff
@@ -128,9 +128,10 @@ def compile_and_run(device, reset_torch_dynamo, request):
             end = time.perf_counter() * 1000
             comp_runtime_metrics = {"success": True, "run_time": round(end - start, 2)}
             option._out_fx_graphs[0].print_tabular()
-            accuracy = calculate_accuracy(outputs, outputs_after)
-            if accuracy:
-                comp_runtime_metrics["accuracy"] = accuracy
+            if model_name not in ["speecht5-tts"]:
+                accuracy = calculate_accuracy(outputs, outputs_after)
+                if accuracy:
+                    comp_runtime_metrics["accuracy"] = accuracy
```
swimdi, line 134: Because speecht5_tts length is dependent, and there has [ttnn.add accuracy issue](https://github.com/tenstorrent/pytorch2.0_ttnn/issues/393) cause the sample result length not equal and cause calculate_accuracy failed, which stop the CI, so I filter speecht5_tts out

kevinwuTT (approved):
I think this looks fine. I rebased my [layout branch](https://github.com/tenstorrent/pytorch2.0_ttnn/tree/kw/layout_device_ops) on top of this and I was able to make compatible changes. So similar to https://github.com/tenstorrent/pytorch2.0_ttnn/pull/383, I'll build my changes on top of this if tests and models are passing.

Issue #391, opened 2024-11-07
Suggest ttnn.to_torch can specify dtype


swimdi:
Because the [problem described](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/383) in there, so there always followed by an `aten.to_copy.default`

And because there will [at least remain an aten.to_copy](https://github.com/tenstorrent/pytorch2.0_ttnn/pull/383#discussion_r1832022697), so cannot end-to-end compilation

But if `ttnn.to_torch` can specify dtype, then can fuse `aten.to_copy.default` on it and can solve the problem.

swimdi:
Or if I skip adding `aten.to_copy.default` for the output type is float/bfloat16 case?

falcon needs `aten.to_copy.default` for the third selection of `aten.index`, so falcon keep `aten.to_copy.default` if just skip the float/bfloat16 case

PR #401, opened 2024-11-11
Use ttnn.typecast to replace aten._to_copy(dtype)

swimdi:
### Ticket
#391 

### Problem description
Current compiler at least remain one `aten._to_copy` in compiled graph and then it never reach the end to end compiled, @pschang-phy suggest using `ttnn.typecast` instead of `aten.to_copy`

### What's changed
using `ttnn.typecast` instead of `aten.to_copy`


swimdi:
cannot keep torch.bool and there are symInt issue appeared, so close this PR

PR #414, opened 2024-11-14
Convert `aten.convolution` to `ttnn.conv2d` or `ttnn.conv_transpose2d`

jerrysky3:
### Ticket
Link to Github Issue

### Problem description

### What's changed


torch_ttnn/passes/lowering/target_wrappers.py
```diff
@@ -16,3 +16,71 @@ def repeat(t, sizes):
 @torch.fx.wrap
 def pack_to_tuple(*args):
     return tuple(args)
+
+
+@torch.fx.wrap
+def move_to_host(device_tensor, layout):
+    host_tensor = ttnn.from_device(device_tensor)
+    return ttnn.to_layout(host_tensor, layout)
```
ayerofieiev-tt, line 24: Means layout change is on host. In some cases its possible to do layout change on device.
Also, I am not sure if plan to_layout removes tile padding.

torch_ttnn/passes/lowering/target_wrappers.py
```diff
@@ -16,3 +16,71 @@ def repeat(t, sizes):
...
+    groups,
+    device,
+    transposed,
+    output_padding=None,
+):
+    if transposed:
+        conv_config = ttnn.Conv2dConfig(
+            shard_layout=ttnn.TensorMemoryLayout.HEIGHT_SHARDED,
+        )
+        output_tensor, _, _, _, _ = ttnn.conv_transpose2d(
```
ayerofieiev-tt, line 50: just fyi, I think other outputs are pre-processed inputs, which can be reused when more convs are called on same inputs.

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -934,6 +933,82 @@ def rewrite_node(node):
...
+                if groups > 1:
+                    return None
+
+                input_tensor = insert_nchw_to_nhwc(g, input_node)
+                # TODO(tt-metal#15148): ttnn.conv2d internal reshape fails with padding
+                input_tensor = g.call_function(ttnn.reshape, (input_tensor, (1, 1, batch_size * in_h * in_w, in_c)))
+                weight_tensor = g.call_function(
+                    target_wrappers.move_to_host,
+                    (weight_node, TtnnRowMajorLayout()),
+                )
```
ayerofieiev-tt, line 978: two things:
* ideally we don't put it to device in the first place. would be nice if there is a pass which can eliminate this to_device, from_device pair
* ideally we store and reuse preprocessed weights tensor produced by conv (there is a PR in metal which moves preprocessing into a separate OP to give more control over it

Nothing to be done in this PR, but want to share this notes as opportunity for further joint improvement.

ayerofieiev-tt (approved):

torch_ttnn/passes/lowering/to_tt_pass.py
jerrysky3: Agree. `torch_ttnn/passes/lowering/eliminate_coreops_pass.py` would be potential pass to eliminate `to_device` and `move_to_host` when possible.

Have an ttnn op to allow pre-process weight separately will definitely help to cache and reuse pre-processed weight tensor

torch_ttnn/passes/lowering/target_wrappers.py
jerrysky3: IMO it will be eaiser to pre-process inputs and cache them if ttnn can expose API to do that instead of doing inside `ttnn.conv2d` automatically, so we can control when to pre-process and cache the inputs

If such separation is not possible in short term and we need to start optimizing for performance, then we can try to cache the other outputs here and use them later.

ayerofieiev-tt: Such separation is possible. The PR to enable that is in progress. I will inform once it is completed.

jerrysky3:
Currently blocked by https://github.com/tenstorrent/pytorch2.0_ttnn/pull/488#issuecomment-2492909088

jerrysky3:
Regenerate the guards with the latest tt-metal version `0.53.1rc1` and filed the bug #15482 for the new issue with the current version

PR #455, opened 2024-11-20
Enable torch.ops.aten.cat.default lowering to ttnn.concat

pschang-phy:
### Ticket
N/A

### Problem description

The concat op `torch.ops.aten.cat.default` should be lowering to TTNN.

### What's changed

`torch.ops.aten.cat.default` lowering to `ttnn.concat`


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -863,6 +863,13 @@ def rewrite_node(node):
                 output_4d = g.call_function(ttnn.moreh_cumsum, (input_4d, dim), kwargs)
                 return g.call_function(ttnn.reshape, (output_4d, input_shape))
 
+            if node.target == torch.ops.aten.cat.default:
+                tensors, dim = args
+                tensor_list = []
+                for tensor in tensors:
+                    tensor_list.append(g.call_function(ttnn.to_layout, (tensor,), {"layout": TtnnTileLayout()}))
```
ayerofieiev-tt, line 870: Can you share more about why this is needed?

pschang-phy: The additional `ttnn.to_layout` is to ensure that all input tensors will be moved to TTNN device, after the `AddDataMovePass‎`.

#### Why not implement in `AddDataMovePass‎`? 
The **args[0]** of `ttnn.concat` is a **list of tensors**, rather than a node type that currently `AddDataMovePass‎` identify to insert `ttnn.from_torch`.
Hence, I prefer to insert `ttnn.to_layout` here in advance acting as mark of data movement, rather coding a special handle in `AddDataMovePass‎`

ayerofieiev-tt: Can you share why existing passes did not make sure that tensor is on device at this moment?
My initial reaction is that it must be handled in a corresponding pass

pschang-phy: Currently`AddDataMovePass`‎ should responsible for data moving in/out.

- Checks in here
The `ttnn.concat` expects an argument of type **List**, causing it to fail the check
https://github.com/tenstorrent/pytorch2.0_ttnn/blob/bedf7b60f42f0d39f30c754a8bd1dd14e58a3ddb/torch_ttnn/passes/lowering/add_data_move_pass.py#L234-L236

- And add data moving node here
For argument of **List** type, adding moving node should revise to consider all tensors in List (eg. `ttnn.concat`)
https://github.com/tenstorrent/pytorch2.0_ttnn/blob/bedf7b60f42f0d39f30c754a8bd1dd14e58a3ddb/torch_ttnn/passes/lowering/add_data_move_pass.py#L311-L325


I think for this specific `ttnn.concat` op, it's more concise to add `ttnn.to_layout` in lowing stage, and let `AddDataMovePass`   insert data moving node for `ttnn.to_layout`.

pschang-phy: the input tensors of `ttnn.concat` are not all guarantee exist in device, there are 2 cases
1. tensor is model inputs
2. tensor is from torch's op

I have commit a test case, as below
```bash
pytest -s tests/lowering/tensor_manipulation/test_concat.py
```

It runs on simple graph
```python
    def forward(self, x1, x2, dim):
        out_relu = torch.ops.aten.relu.default(x2)
        return torch.cat((x1, out_relu), dim)
```

#### compiled result

```
graph():
    %arg0_1 : [num_users=1] = placeholder[target=arg0_1]
    %arg1_1 : [num_users=1] = placeholder[target=arg1_1]
    %ttnn_from_torch : [num_users=1] = call_function[target=ttnn.decorators.ttnn.from_torch](args = (%arg0_1,), kwargs = {layout: ttnn_TILE_LAYOUT, device: ttnn_Specified_Device, dtype: ttnn_bfloat16})
    %ttnn_relu : [num_users=1] = call_function[target=ttnn.decorators.ttnn.relu](args = (%ttnn_from_torch,), kwargs = {})
    %ttnn_from_torch_1 : [num_users=1] = call_function[target=ttnn.decorators.ttnn.from_torch](args = (%arg1_1,), kwargs = {layout: ttnn_TILE_LAYOUT, device: ttnn_Specified_Device, dtype: ttnn_bfloat16})
    %ttnn_to_layout : [num_users=1] = call_function[target=ttnn.decorators.ttnn.to_layout](args = (%ttnn_from_torch_1,), kwargs = {layout: ttnn_TILE_LAYOUT})
    %ttnn_to_layout_1 : [num_users=1] = call_function[target=ttnn.decorators.ttnn.to_layout](args = (%ttnn_relu,), kwargs = {layout: ttnn_TILE_LAYOUT})
    %ttnn_concat : [num_users=1] = call_function[target=ttnn.decorators.ttnn.concat](args = ([%ttnn_to_layout, %ttnn_to_layout_1], 1), kwargs = {})
    %ttnn_to_torch : [num_users=1] = call_function[target=ttnn.decorators.ttnn.to_torch](args = (%ttnn_concat,), kwargs = {})
    return (ttnn_to_torch,)
```

In spite of the insertion of `ttnn.to_layout` in advance will cause a redundant `ttnn.to_layout` layer on `%ttnn_relu` tensor, it will guarantee all tensors (including model input `%arg0_1`) used by `ttnn.concat` will exist in device.
```
%ttnn_to_layout_1 : [num_users=1] = call_function[target=ttnn.decorators.ttnn.to_layout](args = (%ttnn_relu,), kwargs = {layout: ttnn_TILE_LAYOUT})
```

And, the latter `AddDataMovePass` will ‎responsible for data move in when encounter `ttnn.to_layout` op, that's why guarantee all tensors exist in device before using by `ttnn.concat`

ayerofieiev-tt (approved):

PR #498, opened 2024-11-25
Fix issues cause CI fails

pschang-phy:
### Ticket
N/A

### Problem description
CI fails on

- opencv dependency https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/11966694211/job/33411174595

- ttnn.sum issues https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/11982764756/job/33411292571#step:4:4486
Found the `ttnn.sum` ran into wrong results (also happened in local machine). But after reset board n-150, it backs to normal.

### What's changed
- Try fixing dependency error of yolov5
- Skip `ttnn.sum` related tests temporarily


requirements-dev.txt
```diff
@@ -17,3 +17,4 @@ diffusers
 pytorchcv
 librosa
 soundfile
+opencv-python-headless
```
jerrysky3, line 20: Maybe we can pin the package version? So the package won't change by surprise

tests/lowering/misc/test_if.py
```diff
@@ -19,6 +19,7 @@ def forward(self, x):
     "input_shape",
     [(4, 4)],
 )
+@pytest.mark.skip(reason="Temporarily disabled due to reset issue in ttnn.sum")
```
jerrysky3, line 22: If there is an issue to track this issue, the issue number can be added here so we won't turn them back after the problem is resolved

ayerofieiev-tt: agreed

requirements-dev.txt
ayerofieiev-tt: I am ok to allow a similar problem to appear again before we decide to pin the package version.
Wdyt?

tests/lowering/misc/test_if.py
pschang-phy: I fire this issue https://github.com/tenstorrent/pytorch2.0_ttnn/issues/499 to track

tests/lowering/misc/test_sum.py
```diff
@@ -23,6 +23,7 @@ def forward(self, tensor):
         (1,),
     ],
 )
+@pytest.mark.skip(reason="Temporarily disabled due to issue#499")
```
jerrysky3, line 26:
```suggestion
@pytest.mark.skip(reason="Temporarily disabled due to issue (#499)")
```

jerrysky3 (approved):

Issue #439, opened 2024-11-15
aten.t.default


ayerofieiev-tt:

jerrysky3:
Considered it is done with workaround for now. The follow-up improvement is tracked in #377

ayerofieiev-tt:
@jerrysky3 , please take a look at this comment
https://github.com/tenstorrent/pytorch2.0_ttnn/issues/377#issuecomment-2489911960

Transpose recently got improved and Row Major tensor with odd # of elements on width in bfloat16 is not supported nicely.
Switching to Transpose is an option now.

jerrysky3:
> [@jerrysky3](https://github.com/jerrysky3) , please take a look at this comment [#377 (comment)](https://github.com/tenstorrent/pytorch2.0_ttnn/issues/377#issuecomment-2489911960)
> 
> Transpose recently got improved and Row Major tensor with odd # of elements on width in bfloat16 is not supported nicely. Switching to Transpose is an option now.

Oh I see. I misread the comment. Yes, tested with `ttnn.transpose` and all cases are passed now, will send a PR

jerrysky3:
Sent out the PR to use `ttnn.transpose` https://github.com/tenstorrent/pytorch2.0_ttnn/pull/500

PR #500, opened 2024-11-25
Use `ttnn.transpose` for aten transpose ops

jerrysky3:
### Ticket
#439

### Problem description
`ttnn.transpose` has been improved to support all use cases of `aten.t.default` and `aten.tranpose.int`. Use it instead of `ttnn.permute`

### What's changed
- Convert `aten.t.default` and `aten.tranpose.int` into `ttnn.tranpose`
- Enable all tests

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -707,12 +707,12 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool
                 return g.call_function(ttnn.reshape, args=(args[0], output_size))
 
             if node.target in [torch.ops.aten.transpose.int, torch.ops.aten.t.default]:
+                rank = len(node.meta["val"].size())
+                if rank < 2:
+                    # Less 2D transpose is no-op
+                    return args[0]
                 if node.target == torch.ops.aten.t.default:
-                    rank = len(node.meta["val"].size())
-                    assert rank >= 0 and rank <= 2, "Input tensor can only be 0D, 1D or 2D"
-                    if rank < 2:
-                        # Less 2D transpose is no-op
-                        return args[0]
+                    assert rank == 2, "Input tensor should only be 2D here"
                     dim0 = 0
                     dim1 = 1
                 else:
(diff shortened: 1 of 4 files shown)


ayerofieiev-tt (approved):

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -719,21 +718,7 @@ def rewrite_node(node):
...
                     dim0 = args[1]
                     dim1 = args[2]
-                permutation = list(range(rank))
-                permutation[dim0], permutation[dim1] = (
-                    permutation[dim1],
-                    permutation[dim0],
-                )
-                # TODO(#377): ttnn.permute fails when swapping inner-most dim = 1 for 2D
-                if rank == 2 and output_size[0] == 1:
-                    return None
```
ayerofieiev-tt, line 729: This might still fail. <2D tensors when converted back to torch might get 2D. The team is working on this right now.

jerrysky3: It is a no-op to transpose a < 2D tensor. We already have this shortcut for `aten.t` before and I extend it for `aten.transpose` too. So we can avoid this problem here

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -707,12 +707,12 @@ def rewrite_node(node):
                 return g.call_function(ttnn.reshape, args=(args[0], output_size))
 
             if node.target in [torch.ops.aten.transpose.int, torch.ops.aten.t.default]:
+                rank = len(node.meta["val"].size())
+                if rank < 2:
+                    # Less 2D transpose is no-op
+                    return args[0]
                 if node.target == torch.ops.aten.t.default:
-                    rank = len(node.meta["val"].size())
                     assert rank >= 0 and rank <= 2, "Input tensor can only be 0D, 1D or 2D"
```
ayerofieiev-tt, line 715: this assert does not make sense anymore, right?

jerrysky3: It is still the assumption given by the torch.t op https://pytorch.org/docs/stable/generated/torch.t.html. I'm fine to remove it if you think it isn't necessary

ayerofieiev-tt: We can't remove, but given the above `if rank < 2:`, we might want to adjust

jerrysky3: I see. Changed the assert to `rank == 2` as it should be the only possible case when reaching this line

Issue #475, opened 2024-11-21
aten._adaptive_avg_pool2d.default


ayerofieiev-tt:

swimdi:
Need to solved https://github.com/tenstorrent/tt-metal/issues/12151 or the `ttnn.avg_pool2d` cannot support case of `output_size != [1, 1]`
```
input = torch.randn([1, 512, 7, 7]).to(torch.bfloat16)
output_size = [7, 7]
adaptive_avg_pool2d = torch.ops.aten._adaptive_avg_pool2d.default(input, output_size)
print(adaptive_avg_pool2d.shape)
# torch.Size([1, 512, 7, 7])
```

swimdi:
Sorry the above comment is wrong, please ignore it

Here is the description of `aten._adaptive_avg_pool2d.default`:

`aten._adaptive_avg_pool2d.default` is not same with `aten.avg_pool2d.default`, but it can [convert to `aten.avg_pool2d.default`](https://stackoverflow.com/questions/58692476/what-is-adaptive-average-pooling-and-how-does-it-work) by mapping `aten._adaptive_avg_pool2d.default`'s `output_size` argument to `aten.avg_pool2d.default`'s `stride` & `size` by

```
Stride = (input_size//output_size)  
Kernel size = input_size - (output_size-1)*stride  
Padding = 0
```

So if want to support lowering of `aten._adaptive_avg_pool2d.default`, then it needs either `ttnn.adaptive_avg_pool2d` op for direct converting or `ttnn.avg_pool2d` with above adjust

But seems there no `ttnn.adaptive_avg_pool2d` op now, and `ttnn.avg_pool2d` is not finished according to https://github.com/tenstorrent/tt-metal/issues/12151

Luckly, there's [only one input variation](https://github.com/tenstorrent/pytorch2.0_ttnn/blob/main/docs/operations/aten._adaptive_avg_pool2d.default.md) appeared on models
```
Tensor<[1, 512, 7, 7]> self = ?, List[int] output_size = [7, 7]
```
And this input variation actually let `aten._adaptive_avg_pool2d.default` do nothing, so if we check these input varation type during lowering and just remove it, then we can say `aten._adaptive_avg_pool2d.default` is fully supported of the current scope of models

PR #510, opened 2024-11-26
Fix lowering of aten._adaptive_avg_pool2d.default

swimdi:
### Ticket
#475 

### Problem description
Current lowering of `aten._adaptive_avg_pool2d.default` is wrong, and `ttnn.avg_pool2d` is not finished according to https://github.com/tenstorrent/tt-metal/issues/12151, so disable its lowering

Additionally, some special input varation let `aten._adaptive_avg_pool2d.default` do nothing, add the code to detect it and bypass it

### What's changed
 - Disable lowering of `aten._adaptive_avg_pool2d.default`
 - Check special input variation pattern of aten._adaptive_avg_pool2d.default
 - Remove blocklist of `aten._adaptive_avg_pool2d.default`

### DIFF
--- a/tests/pattern/test_vgg_pattern.py
+++ b/tests/pattern/test_vgg_pattern.py
@@ -1,0 +1,27 @@
+import torch
+import torch_ttnn
+import pytest
+import ttnn
+
+
+class PatternModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, input, output_size):
+        return torch.ops.aten._adaptive_avg_pool2d.default(input, output_size)
+
+
+def test_vgg_adaptive_avg_pool2d(device):
+    m = PatternModule()
+    input = torch.randn([1, 512, 7, 7]).to(torch.bfloat16)
+    output_size = [7, 7]
+    result_before = m.forward(input, output_size)
+    option = torch_ttnn.TorchTtnnOption(device=device)
+    m = torch.compile(m, backend=torch_ttnn.backend, options=option)
+    result_after = m.forward(input, output_size)
+    assert result_before.shape == result_after.shape
+    assert torch.allclose(result_before, result_after, rtol=0.1, atol=0.1)
+    assert not any(
+        node.target == torch.ops.aten._adaptive_avg_pool2d.default for node in option._out_fx_graphs[0].nodes
+    )
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -318,14 +318,12 @@ class ReplaceMoreTt(torch.fx.Transformer):
                     return True
                 return False
 
-            if hasattr(args[0], "node") and args[0].node.name in self._input_node_meta:
-                arg0_meta = self._input_node_meta[args[0].node.name]
-            else:
-                arg0_meta = None
-            if hasattr(args[1], "node") and args[1].node.name in self._input_node_meta:
-                arg1_meta = self._input_node_meta[args[1].node.name]
-            else:
-                arg1_meta = None
+            arg0_meta = None
+            arg1_meta = None
+            if hasattr(args[0], "node") and hasattr(args[0].node, "meta"):
+                arg0_meta = args[0].node.meta
+            if hasattr(args[1], "node") and hasattr(args[1].node, "meta"):
+                arg1_meta = args[1].node.meta
             if is_zero_dim(arg0_meta) or is_zero_dim(arg1_meta):
                 return self.call_function_prop_meta(target, args, kwargs)
             return self.call_function_prop_meta(ttnn.add, args, kwargs)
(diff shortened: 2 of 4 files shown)


tests/pattern/test_vgg_pattern.py
```diff
@@ -0,0 +1,25 @@
...
+def test_vgg_adaptive_avg_pool2d(device):
+    m = PatternModule()
+    input = torch.randn([1, 512, 7, 7]).to(torch.bfloat16)
+    output_size = [7, 7]
+    result_before = m.forward(input, output_size)
+    option = torch_ttnn.TorchTtnnOption(device=device)
+    m = torch.compile(m, backend=torch_ttnn.backend, options=option)
+    result_after = m.forward(input, output_size)
+    assert result_before.shape == result_after.shape
+    assert torch.allclose(result_before, result_after, rtol=0.1, atol=0.1)
```
jerrysky3, line 24: Maybe we need to check if the op is converted? E.g.

```py
    # Check the graph has be rewritten and aten ops are replaced
    assert not any(node.target == torch.ops.aten._adaptive_avg_pool2d.default for node in option._out_fx_graphs[0].nodes))
```

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -400,8 +404,22 @@ def is_zero_dim(meta):
...
+                output_size = kwargs["output_size"]
+            new_graph_node_meta = self.get_new_graph_node_meta()
+            if hasattr(args[0], "node") and args[0].node.name in new_graph_node_meta:
+                arg0_meta = new_graph_node_meta[args[0].node.name]
+                arg0_size = list(arg0_meta["val"].size())
+            # special case: if output_size is same with input last two dimensions
+            # then actually this op do nothing
+            if output_size is not None and arg0_size is not None and arg0_size[-2:] == output_size:
+                return args[0]
+            # TODO: no ttnn op can convert _adaptive_avg_pool2d
```
jerrysky3, line 422: `ttnn.global_avg_pool2d` actually can handle the case of output_size `(1, 1)` I guess

But I'm fine to convert it now (which is not used) or later with the actual `ttnn.avg_pool2d` implemented

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -113,6 +113,9 @@ def placeholder(self, target, args, kwargs):
             proxy.node.meta = self._input_node_meta[proxy.node.name]
         return proxy
 
+    def get_new_graph_node_meta(self):
+        return {node.name: node.meta for node in self.new_graph.nodes}
```
jerrysky3, line 117: Where does the `self.new_graph` come from?

swimdi: I think the mechanism of `torch.fx.Transformer` is one by one transom origin graph's op to `new_graph`
https://pytorch.org/docs/stable/fx.html#torch.fx.Transformer
Here's its implement code
https://pytorch.org/docs/stable/_modules/torch/fx/interpreter.html#Transformer

The problem I encounter here is:
 original graph have 8 getitem
```
getitem0 = max_pool_0[0]
getitem1 = max_pool_0[1]
getitem2 = max_pool_1[0]
getitem3 = max_pool_1[1]
...
```
And the input of `adaptive_avg_pool` is
```
adaptive_avg_pool_0.args[0] = getitem0
adaptive_avg_pool_1.args[0] = getitem2
```
And `self._input_node_meta`  record these original graph's information

But the `new_graph`, due to `getitem1` & `getitem3` has no user, so maybe `remove_unused_op` method is applied, so the `new_graph` is
```
getitem0 = max_pool_0[0]
getitem1 = max_pool_1[0]
adaptive_avg_pool_0.args[0] = getitem0
adaptive_avg_pool_1.args[0] = getitem1
```
So in the lowering of `adaptive_avg_pool`, these code is wrong
```
# in adaptive_avg_pool_1
input_name = node.args[0].name # getitem1, it actually from max_pool_1[0]
input_shape = self._input_node_meta[input_name] # access origin graph and get max_pool_0[1], which is wrong!
```

So cannot use `self._input_node_meta`, use `get_new_graph_node_meta` instead

And because `new_graph` is generting at this moment, so it cannot set at `__init__`, instead use `get_new_graph_node_meta` to get the real time updates

tests/pattern/test_vgg_pattern.py
swimdi: okok, I had added this

torch_ttnn/passes/lowering/to_tt_pass.py
swimdi: Updated:
After discuss with @jerrysky3 , turns out just use `node.args[0].node.meta` can satisfy the requiremt, so no need the `get_new_graph_node_meta` function

jerrysky3 (approved):

swimdi:
Hi @ayerofieiev-tt is this PR ok to merge?

ayerofieiev-tt (approved):

PR #513, opened 2024-11-26
Move blocklist of `aten.constant_pad_nd.default` to to_tt_pass.py

swimdi:
### Ticket
#478 

### Problem description
Fix check of lowering `aten.constant_pad_nd.default` so that can remove its blocklist, now these situation will runtime failed so can just fallback
1. rank > 4:
will TT_THROW `Tensor rank is greater than 4` (btw, there's no rank>4 case in models, so can lower its priority)
2. output_shape[-1] % 2 != 0:
will TT_THROW `pad_op.cpp:34: this->output_tensor_shape[3] % 2 == 0  RM padding requires output X dim to be a multiple of 2`
3. any(p < 0 for p in pad): i.e., pad cannot < -1
will `TypeError: __call__(): incompatible function arguments. The following argument types are supported: ...`

### What's changed
 - Move blocklist of aten.constant_pad_nd.default to to_tt_pass.py


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -745,6 +751,8 @@ def rewrite_node(node):
                     or full_pad[-2][1] % ttnn.TILE_SIZE != 0
                 ):
                     input = g.call_function(ttnn.to_layout, args=(input, TtnnRowMajorLayout()))
+                if output_shape[-1] % 2 != 0:
```
ayerofieiev-tt, line 754: This must be very easy to fix now that we allow odd number values on the innermost  dimension in bfloat16

ayerofieiev-tt:
Please fire us tickets for each individual error.

ayerofieiev-tt (approved):

PR #518, opened 2024-11-27
Fix `aten.masked_fill.Scalar` and `aten.masked_fill.Tensor` lowering

pschang-phy:
### Ticket
https://github.com/tenstorrent/pytorch2.0_ttnn/issues/483

### Problem description
`aten.masked_fill.Scalar` and `aten.masked_fill.Tensor` fail to lowering into TTNN ops.

### What's changed
1. Fix the `ttnn.full` usage, the fill value should be set in `kwargs["fill_value"]`
2. extract fill value from `getattr` constant in `aten.masked_fill.Tensor` argument 

Test
```bash
pytest -s tests/lowering/creation/test_masked_fill.py
```


ayerofieiev-tt (approved):

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -828,17 +829,28 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
...
                 np_mask_shp = np.array(mask_shape)
                 if np.sum(np_tensor_shp % np_mask_shp) == 0:
-                    multiplier = np_tensor_shp // np_mask_shp
-                    mask_bcst = g.call_function(target_wrappers.repeat, args=(mask, multiplier.tolist()))
+                    if isinstance(fill_value, (float,)):
+                        fill_real_value = fill_value
+                    elif fill_value.op == "get_attr":
+                        with unset_fake_temporarily():
+                            fill_value_attr = getattr(gm, fill_value.target)
+                            fill_real_value = fill_value_attr.item()
```
jerrysky3, line 837: Had an offline discussion. This is to handle the special case when `masked_fill.Tensor` is called with a 0-D fill value tensor. Maybe we can add some comments here and guard unsupported cases

swimdi:
There have `aten_masked_fill_scalar_blocklist` in `to_tt_guard.py` & `to_tt_guard_autogen.py` I think you could remove

torch_ttnn/passes/lowering/to_tt_pass.py
swimdi: I think the unsupported cases should fire an issue for each so that can show on the roadmap

pschang-phy: I have revised to more general condition check 54f20ca107a3119627d471ce5d399d7105444f2a for all case of constant, support either scalar or `torch.tensor`. Also add comment to mention this.

For all our current tests, not encounter unsupported cases. Furthermore, it seems in `torch.masked_fill` API not support tensor fill value (only support scalar) directly, wondering how this **distilbert-base-uncased** model got this `masked_fill.Tensor` op

tests/lowering/creation/test_masked_fill.py
jerrysky3, line 1: Maybe we can also add test cases for the current unsupported cases and make sure they fallback correctly

jerrysky3 (approved):

tests/lowering/creation/test_masked_fill.py
pschang-phy: it seems in `torch.masked_fill` API not support input with **tensor fill value** (only support scalar) directly.

PR #522, opened 2024-11-28
Enable `torch.ops.aten.slice_scatter.default`  lowering

pschang-phy:
### Ticket
https://github.com/tenstorrent/pytorch2.0_ttnn/issues/485

### Problem description
`torch.ops.aten.slice_scatter.default`  remains not lowering to TTNN.

### What's changed
`torch.ops.aten.slice_scatter.default(input, src, dim, start, end)` could be lowered into `ttnn.slice` + `ttnn.concat`
or
just return `src` tensor, if size of `src` tensor == size of `input` tensor

The only 2 models, **ghostnet_100.in1k** and **ghostnetv2_100.in1k** which has `torch.ops.aten.slice_scatter.default` op are the case of just return `src` tensor

Test
```bash
pytest -s tests/lowering/tensor_manipulation/test_slice_scatter.py
```

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -1060,7 +1060,8 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool
 
                 # slice_scatter could be concat([pre_slice_tensor, src_tensor, post_slice_tensor])
                 rank = len(tensor_shape)
-                end = min(tensor_shape[dim], end) if end >= 0 else (end + tensor_shape[dim])
+                end = end if end >= 0 else (end + tensor_shape[dim])
+                end = 0 if end < 0 else min(tensor_shape[dim], end)
                 [step] = step or [1]
 
                 if step != 1 or dim >= rank:
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -1047,6 +1047,49 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool
                 output_shape = node.meta["val"].size()
                 return insert_sharded_nhwc_to_nchw(g, output_tensor, output_shape)
 
+            if node.target == torch.ops.aten.slice_scatter.default:
+                tensor, src_tensor, dim, start, end, *step = args
+
+                tensor_shape = tensor.meta["val"].size()
+                src_tensor_shape = src_tensor.meta["val"].size()
+
+                # nothing to slice, result tensors == src_tensor
+                if tensor_shape == src_tensor_shape:
+                    return src_tensor
+
+                # slice_scatter could be concat([pre_slice_tensor, src_tensor, post_slice_tensor])
+                rank = len(tensor_shape)
+                end = min(tensor_shape[dim], end) if end >= 0 else (end + tensor_shape[dim])
+                [step] = step or [1]
+
+                if step != 1 or dim >= rank:
+                    return None
+
+                tensors_list = []
+                # pre_slice_tensor
+                if start > 0:
+                    slice_start = np.zeros(rank, dtype=int)
+                    slice_end = list(tensor_shape)
+                    slice_end[dim] = start
+                    tensors_list.append(g.call_function(ttnn.slice, (tensor, [*slice_start], [*slice_end])))
+
+                # src_tensor
+                tensors_list.append(src_tensor)
+
+                # post_slice_tensor
+                if end < tensor_shape[dim]:
+                    slice_start = np.zeros(rank, dtype=int)
+                    slice_start[dim] = end
+                    slice_end = list(tensor_shape)
+                    tensors_list.append(g.call_function(ttnn.slice, (tensor, [*slice_start], [*slice_end])))
+
+                # concat all together
+                tensors_to_concat = []
+                for tensor in tensors_list:
+                    tensors_to_concat.append(g.call_function(ttnn.to_layout, (tensor,), {"layout": TtnnTileLayout()}))
+
+                return g.call_function(ttnn.concat, (tensors_to_concat, dim))
+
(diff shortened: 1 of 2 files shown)


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1048,6 +1048,49 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
...
+                tensor_shape = tensor.meta["val"].size()
+                src_tensor_shape = src_tensor.meta["val"].size()
+
+                # nothing to slice, result tensors == src_tensor
+                if tensor_shape == src_tensor_shape:
+                    return src_tensor
+
+                # slice_scatter could be concat([pre_slice_tensor, src_tensor, post_slice_tensor])
+                rank = len(tensor_shape)
+                end = min(tensor_shape[dim], end) if end >= 0 else (end + tensor_shape[dim])
```
jerrysky3, line 1063: The `end` might still be out of range if it is too small on the negative side?

```suggestion
                end = end if end >= 0 else (end + tensor_shape[dim])
                end = 0 if end < 0 else min(tensor_shape[dim], end)
```

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1048,6 +1048,49 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
...
+                [step] = step or [1]
+
+                if step != 1 or dim >= rank:
```
jerrysky3, line 1066: Is `dim >= rank` supported by `torch.slice_scatter`? If not maybe we should use assert to fail fast here?

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1048,6 +1048,49 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
...
+                    return None
+
+                tensors_list = []
+                # pre_slice_tensor
+                if start > 0:
+                    slice_start = np.zeros(rank, dtype=int)
```
jerrysky3, line 1072: IMO `[0] * rank` is a little bit simpler. I'm fine with both ways

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1048,6 +1048,49 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
...
+                    slice_end = list(tensor_shape)
+                    slice_end[dim] = start
+                    tensors_list.append(g.call_function(ttnn.slice, (tensor, [*slice_start], [*slice_end])))
+
+                # src_tensor
+                tensors_list.append(src_tensor)
+
+                # post_slice_tensor
+                if end < tensor_shape[dim]:
+                    slice_start = np.zeros(rank, dtype=int)
```
jerrysky3, line 1082: Same here

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1048,6 +1048,49 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
...
+                if step != 1 or dim >= rank:
+                    return None
+
+                tensors_list = []
+                # pre_slice_tensor
+                if start > 0:
+                    slice_start = np.zeros(rank, dtype=int)
+                    slice_end = list(tensor_shape)
+                    slice_end[dim] = start
+                    tensors_list.append(g.call_function(ttnn.slice, (tensor, [*slice_start], [*slice_end])))
```
jerrysky3, line 1075: If both `slice_start` and `slice_end` are list, we can directly pass them?

```suggestion
                    tensors_list.append(g.call_function(ttnn.slice, (tensor, slice_start, slice_end)))
```

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1048,6 +1048,49 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
...
+
+                # src_tensor
+                tensors_list.append(src_tensor)
+
+                # post_slice_tensor
+                if end < tensor_shape[dim]:
+                    slice_start = np.zeros(rank, dtype=int)
+                    slice_start[dim] = end
+                    slice_end = list(tensor_shape)
+                    tensors_list.append(g.call_function(ttnn.slice, (tensor, [*slice_start], [*slice_end])))
```
jerrysky3, line 1085: Same here

jerrysky3 (approved):

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1048,6 +1048,49 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
                 output_shape = node.meta["val"].size()
                 return insert_sharded_nhwc_to_nchw(g, output_tensor, output_shape)
 
+            if node.target == torch.ops.aten.slice_scatter.default:
+                tensor, src_tensor, dim, start, end, *step = args
```
swimdi, line 1052: [start & end seems optional in there](https://pytorch.org/docs/stable/torch.compiler_ir.html), and maybe appeared in kwargs, maybe should have a check

torch_ttnn/passes/lowering/to_tt_pass.py
pschang-phy: ok, I have changed to assert back early.

torch_ttnn/passes/lowering/to_tt_pass.py
pschang-phy: changed to `[0] * rank`, it's more simpler

torch_ttnn/passes/lowering/to_tt_pass.py
pschang-phy: I have revised it.

torch_ttnn/passes/lowering/to_tt_pass.py
pschang-phy: I have revised to more check on `start` and `end`

PR #532, opened 2024-12-01
Refactor add data move pass

swimdi:
### Ticket
N/A

### Problem description
Although this pass named `add_data_move`, but put things in perspective, if each node can ensure its input correct, then it also ensure the input is added the appropriate data movement node

So in there introduce `NodeInputAligner` class, which help node's input  **`align`** to the correct posture

In torch_ttnn compiler, the main task of `NodeInputAligner` to deal with has these scope

1. torch to ttnn: key action is `from_torch`
2. ttnn to torch: key action is `to_torch`

worth noting, some ttnn node may needs different layout & device, so also have this scope

3. ttnn to ttnn: key action is `to_layout` and (`from_device` or `to_device`)

for example, if reshape input is `tile_layout`,then the input needs be aligned using `_align_for_special_layout`
```
input_node => to_layout(row_major_layout) => reshape
```
Additionally, since the layout changes to row_major_layout after the reshape operation—different from the default layout—the next node must reset to the default using `_reset_to_default_layout`
```
input_node => to_layout(row_major_layout) => reshape => to_layout(tile_layout)
```
### What's changed
 - Introduce `NodeInputAligner` class to align the node's input as the correct posture
   - Give `AlignSpec*` class for prepare the aligner information and let it be the key of created_aligned_node for reuse
   - Implemented core functions:
     - `_get_align_spec`: Retrieves alignment specifications
     - `_create_aligned_node`: Creates nodes for alignment
     - `_connect_aligned_node`: Connects aligned nodes
   - Added support functions:
     - `_align_for_special_layout`: Handles special layout requirements
     - `_reset_to_default_layout`: Resets to the default layout when necessary

Issue #420, opened 2024-11-15
aten.view's err value too large on integer type


swimdi:
In `microsoft/beit-base-patch16-224`, there has this pattern
```
view_13 = torch.ops.aten.view.default(arg223_1, [-1])
index = torch.ops.aten.index.Tensor(arg1_1, [view_13])
```
where `arg1_1.shape = [732, 12]` and `arg223_1` will to be a list of index
```
tensor([[731, 729, 729,  ..., 729, 729, 729],
        [730, 364, 363,  ...,   2,   1,   0],
        [730, 365, 364,  ...,   3,   2,   1],
        ...,
        [730, 726, 725,  ..., 364, 363, 362],
        [730, 727, 726,  ..., 365, 364, 363],
        [730, 728, 727,  ..., 366, 365, 364]])
```
`arg223_1` dtype is `int64`, and value 731 is contained in it, and
 - if `arg223_1` become `view_13` by `torch.ops.aten.view.default`, then value 731 retain 731
 - if `arg223_1` become `view_13` by `ttnn.reshape`, then because `ttnn.reshape` input dtype must be bfloat16, then value 731 should be converted to `bfloat16(731) = 732`, and convert back to `int64(732) = 732` after `ttnn.reshape`, so the value 731 become 732 in `view_13`, then it will casue this error
```
IndexError: index 732 is out of bounds for dimension 0 with size 732
```

One of the suggestion of this issue is NOT lowering `torch.ops.aten.view.default` to `ttnn.reshape` if dtype is `int64`(then beit cannot end-to-end compiled), but I want to get some feedback from TT side

You can reproduce this error by `pytest tests/pattern/test_beit_pattern.py`

ayerofieiev-tt:
Got it. Checking with the team. If we truly can’t support int64 in reshape, I agree, we should avoid.

ayerofieiev-tt:
@swimdi the team reports that this op must support integer types no problem

Please consult with @kevinwuTT if you want to try maintaining the dtype.
I think this issue must remain on Skymizer.

@ntarafdar FYI

swimdi:
Afterward I found this reason maybe is `add_data_move_pass.py` force `ttnn.from_torch` with `bfloat16` dtype before `ttnn.reshape`

I think I will refactor the `add_data_move_pass.py` and resolved this force convert bfloat16 dtype issue

[...]

PR #533, opened 2024-12-01
Fix reshape value error too large

swimdi:
### Ticket
#420 

### Problem description
 - from_torch not as bfloat16 if val type is int and its user is reshape

### What's changed
 - from_torch not as bfloat16 if val type is int and its user is reshape

I'll merge #532 first then this PR


swimdi:
This CI failed due to [this](https://github.com/tenstorrent/pytorch2.0_ttnn/issues/420#issuecomment-2514202695)

swimdi:
> This CI failed due to https://github.com/tenstorrent/pytorch2.0_ttnn/issues/420#issuecomment-2514202695

Solved at #651 , but seems the yolos problem is also solved now (maybe solved by constant folding), so I can merge this PR first

jerrysky3 (approved):

PR #554, opened 2024-12-04
Add `aten.arange.default` and `aten.view.default` constant folding

pschang-phy:
### Ticket
https://github.com/tenstorrent/pytorch2.0_ttnn/issues/534

### Problem description
`aten.arange.default` op remains not lowering to TTNN.

### What's changed
In current encountered models, `aten.arange.default` op all have constant input, eg.
```
%arange_2 : [num_users=1] = call_function[target=torch.ops.aten.arange.default](args = (128,), kwargs = {dtype: torch.float32, device: cpu, pin_memory: False})
```

It is suitable to constant folding out it.
Add `aten.arange.default` and `aten.view.default` in constant folding.


torch_ttnn/passes/constant_folding_pass.py
```diff
@@ -27,7 +29,7 @@ def call(self, gm: torch.fx.GraphModule):
 
     def _can_fold(self, gm: torch.fx.GraphModule, node):
         for arg in node.args:
-            if isinstance(arg, torch.fx.Node):
+            if not isinstance(arg, (int,)) and isinstance(arg, torch.fx.Node):
```
swimdi, line 32: Maybe also float?

swimdi (approved):

torch_ttnn/passes/constant_folding_pass.py
pschang-phy: yes, added in e689c4ec22690079a08630854c7b5de93f3e2fa3.

Issue #555, opened 2024-12-04
Some group norm input varation failed


swimdi:
These input varations of group_norm are failed
```
"input_shape, num_groups",
# These two cases appeared in stable diffusion v2 and accuracy failed
pytest.param((1, 1280, 8, 8), 32, marks=pytest.mark.xfail(reason="")),
pytest.param((1, 2560, 8, 8), 32, marks=pytest.mark.xfail(reason="")),
# These four cases appeared in retinanet_resnet50_fpn_v2 and RuntimeError
pytest.param((1, 256, 50, 68), 32, marks=pytest.mark.xfail(reason="")),
pytest.param((1, 256, 25, 34), 32, marks=pytest.mark.xfail(reason="")),
pytest.param((1, 256, 13, 17), 32, marks=pytest.mark.xfail(reason="")),
pytest.param((1, 256, 7, 9), 32, marks=pytest.mark.xfail(reason="")),
```

You can reproduce by
1. remove `aten_native_group_norm_default_blocklist`
2. `pytest tests/lowering/normalization/test_group_norm.py --runxfail`

Btw, if the input_tensor layout set `TtnnTileLayout` then this two case accuracy failed
```
"input_shape, num_groups",
[(1, 320, 32, 32), 32]
[(2, 320, 64, 64), 32]
```

Issue #538, opened 2024-12-02
aten.native_group_norm.default


jdh8:

swimdi:
Hi @ayerofieiev-tt now the #556 cannot merged due to `test_group_norm.py` failed
```
pytest tests/lowering/normalization/test_group_norm.py

FAILED tests/lowering/normalization/test_group_norm.py::test_group_norm[input_shape0-32-True] - AssertionError: 0.7651781854059824
FAILED tests/lowering/normalization/test_group_norm.py::test_group_norm[input_shape1-32-True] - AssertionError: 0.7730704167640353
FAILED tests/lowering/normalization/test_group_norm.py::test_group_norm[input_shape2-32-True] - AssertionError: 0.7757072186311633
FAILED tests/lowering/normalization/test_group_norm.py::test_group_norm[input_shape3-32-True] - AssertionError: 0.7739301248305729
```
And I have the impression that this test can passed before, I don't know why now it failed

And because #556 is implemented based on [test_group_norm.py on tt-metal](https://github.com/tenstorrent/tt-metal/blob/56c3e45de335d8320a055ae6f3d4e83d0bdc0c17/tests/ttnn/unit_tests/operations/test_group_norm.py), so I also check it, and find this case is also failed
```
cd tt-metal
pytest tests/ttnn/unit_tests/operations/test_group_norm.py

FAILED tests/ttnn/unit_tests/operations/test_group_norm.py::test_group_norm_with_height_sharded[num_groups=32-W=32-H=32-C=320-N=1] - AssertionError: 0.6637590850804874
FAILED tests/ttnn/unit_tests/operations/test_group_norm.py::test_group_norm_with_block_sharded_v2_8x8_grid[N=2-C=320-H=64-W=64-num_groups=32-device_params={'l1_small_size': 0}] - AssertionError: 0.7984854165794627
FAILED tests/ttnn/unit_tests/operations/test_group_norm.py::test_group_norm_with_block_sharded_v2_8x8_grid[N=1-C=640-H=1-W=2048-num_groups=32-device_params={'l1_small_size': 0}] - AssertionError: 0.7983983472927303
FAILED tests/ttnn/unit_tests/operations/test_group_norm.py::test_group_norm_with_block_sharded_v2_8x8_grid[N=1-C=640-H=1-W=4096-num_groups=32-device_params={'l1_small_size': 0}] - AssertionError: 0.7984879175666398
[...]

PR #556, opened 2024-12-04
Lowering `aten.native_group_norm.default`

swimdi:
### Ticket
#538 

### Problem description
Support the lowering of `aten.native_group_norm.default`

I implement it accroding to [test_group_norm.py on tt-metal](https://github.com/tenstorrent/tt-metal/blob/56c3e45de335d8320a055ae6f3d4e83d0bdc0c17/tests/ttnn/unit_tests/operations/test_group_norm.py)

Originally I reference the `test_group_norm_with_block_sharded_v2_8x8_grid_tile_layout` part, but some cases accuracy failed
```
"input_shape, num_groups",
[(1, 320, 32, 32), 32]
[(2, 320, 64, 64), 32]
```
And after I [set the input_tensor layout as `TtnnRowMajorLayout`](https://github.com/tenstorrent/pytorch2.0_ttnn/blob/320c530b57f649346f9085bf956e3999f0c8afa5/torch_ttnn/passes/lowering/add_data_move_pass.py#L348-L349), its accuracy pass

And there still some case failed, I issued at #555 

### What's changed
 - Support lowering `aten.native_group_norm.default`
 - Add aten_native_group_norm_default_blocklist
 - Add test_group_norm.py

btw, only `stable diffusion v2` & `retinanet_resnet50_fpn_v2` use group_norm, so maybe this op's priority can lower

I'll merge #532 first and then this PR


ayerofieiev-tt:
>And after I [set the input_tensor layout as TtnnRowMajorLayout](https://github.com/tenstorrent/pytorch2.0_ttnn/blob/320c530b57f649346f9085bf956e3999f0c8afa5/torch_ttnn/passes/lowering/add_data_move_pass.py#L348-L349), its accuracy pass

Please fire a ticket. There should be no requirement to change from Tile to RM for PCC to pass. Sounds like a bug to me.

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -499,6 +500,57 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
...
+                N, C, H, W = arg0_shape
+                input_tensor = args[0]
+                weight_tensor = args[1]
+                bias_tensor = args[2]
+                num_groups = args[6]
+                epsilon = args[7]
+                inplace = False
+                grid_size_x = device.compute_with_storage_grid_size().x
+                grid_size_y = device.compute_with_storage_grid_size().y
+                shard_shape = N * H * W // grid_size_x, C // grid_size_y
```
jerrysky3, line 517: The `shaprd_shape` might contains 0 if `N * H * W < grid_size_x` or `C < grid_size_y`. Will that cause problems?

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -499,6 +500,57 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
...
+                )
+
+                output_tensor_l1_reshape = g.call_function(ttnn.reshape, (output_tensor_l1, (N, H, W, C)))
+                output_tensor_l1_permute = g.call_function(ttnn.permute, (output_tensor_l1_reshape, (0, 3, 1, 2)))
+                new_node = output_tensor_l1_permute
+
+                node.replace_all_uses_with(new_node, delete_user_cb=lambda node: node != new_node)
+                node_users = list(new_node.users.keys())
+                for node_user in node_users:
+                    node_user.replace_all_uses_with(new_node)
```
jerrysky3, line 551: What is the reason we need to replace the users of users?

torch_ttnn/utils.py
```diff
@@ -62,3 +67,8 @@ def __repr__(self):
 class TtnnL1MemoryConfig:
     def __repr__(self):
         return f"ttnn_L1_MEMORY_CONFIG"
+
+
+class TtnnCoreGrid:
+    def __repr__(self):
+        return f"ttnn_CORE_GRID"
```
jerrysky3, line 74: This seems to be not used?

torch_ttnn/passes/lowering/to_tt_pass.py
swimdi: Yes this will cause ttnn runtime error, I've add a check to not lowering

torch_ttnn/utils.py
swimdi: Yes it is not used, I've deleted

torch_ttnn/passes/lowering/to_tt_pass.py
swimdi: I referenced from layer_norm

The output of `aten.native_group_norm.default` is (Y, mean, rstd), and the output of `ttnn.layer_norm` is Y, and the original garph is like this
```
group_norm = aten.group_norm
getitem = getitem(group_norm, 0)
return ((getitem,),)
```
and want the graph become this
```
group_norm = ttnn.group_norm
return (group_norm,)
```
so need to replace getitem's uses with group_norm

swimdi:
>> And after I [set the input_tensor layout as TtnnRowMajorLayout](https://github.com/tenstorrent/pytorch2.0_ttnn/blob/320c530b57f649346f9085bf956e3999f0c8afa5/torch_ttnn/passes/lowering/add_data_move_pass.py#L348-L349), its accuracy pass

> Please fire a ticket. There should be no requirement to change from Tile to RM for PCC to pass. Sounds like a bug to me.

that case is mentioned in #555 , or you prefer an individual issue@@?

torch_ttnn/passes/lowering/to_tt_pass.py
jerrysky3: I feel like the issue with this method is that it might go wrong if the graph isn't in the expceted form. For example if the graph calls `getitem(group_norm, 1)` to get mean, then we will mistakenly replace it with `Y` and results in wrong output.

We encountered a similar issue with batch_norm and max_pool2d and here is the solution we used:

https://github.com/tenstorrent/pytorch2.0_ttnn/blob/485dc5b6897faa35922dfb2949b356daf92d7da3/torch_ttnn/passes/lowering/to_tt_pass.py#L482-L491

It contains two parts:

1. `is_getitem_0_only_user` is to check if graph only calls `getitem(x, 0)`. So any unsupported `getitem(x, 1)` will fail-fast.
2. Instead of replacing `getitem(x, 0)` with `x`, it uses `g.call_function(target_wrappers.pack_to_tuple, (output,))` to pack `x` into `(x,)` so the `getitem` later will unpack it and get `x` correctly

IMO 1. is good to have to avoid troubles on debugging incorrect results later

swimdi: okok! This suggestion is great, thank for reminding, I've updated the code

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -499,17 +499,64 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
...
+                input_tensor = args[0]
+                weight_tensor = args[1]
+                bias_tensor = args[2]
+                num_groups = args[6]
+                epsilon = args[7]
+                inplace = False
+                grid_size_x = device.compute_with_storage_grid_size().x
+                grid_size_y = device.compute_with_storage_grid_size().y
+                shard_shape = N * H * W // grid_size_x, C // grid_size_y
+                if shard_shape[0] == 0 or shard_shape[1] == 0:
```
jerrysky3, line 519: Maybe add a TODO to remind us the shard shape computation can be improved later

jerrysky3 (approved):
LGTM on the group norm parts added by this PR

PR #565, opened 2024-12-06
Reorder CI workflow of autogen-op

swimdi:
### Ticket
N/A

### Problem description
For the purpose of saving time, the current workflow `model-tests` and `model-autogen-op-tests` is parallel, so `model-autogen-op-tests` is run the original testscase, not the changed of `push-autogen-op-tests`, so needs to trigger next run to let it adopted
```
# These two job is parallel
model-tests => push-autogen-op-tests (updated test)
model-autogen-op-tests (use original test)
```

But now `model-autogen-op-tests` not very often, just run for `Docs/Tests/All`, so I think it's not necessary for parallel, so maybe let `model-tests` and `model-autogen-op-tests` sequential, and then `model-autogen-op-tests` can run updated case
```
# sequential
model-tests => push-autogen-op-tests (updated test) => model-autogen-op-tests (use updated test)
```

If agree this change, then can merge this PR to let it sequential

### What's changed
 - Reorder the CI workflow


.github/workflows/before_merge.yaml
```diff
@@ -93,8 +93,52 @@ jobs:
           name: model-tests-metrics-group-${{ matrix.group }}
           path: metrics/
 
+  push-autogen-op-tests:
+    needs: [model-tests]
+    if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.commit_report != 'None'}}
```
ayerofieiev-tt, line 98:
```suggestion
    if: ${{ github.event_name == 'workflow_dispatch' && (github.event.inputs.commit_report == 'All' || github.event.inputs.commit_report == 'Tests'}}
```

This will allow to update Docs without regenerating all tests, and will run them in parallel with models, right?

ayerofieiev-tt: also here
```
        description: 'Commit generated report files: None, Docs, All'
```
must update to
```
        description: 'Commit generated report files: None, Docs, Tests, All'
```

.github/workflows/before_merge.yaml
```diff
@@ -93,8 +93,52 @@ jobs:
...
+            exit 0
+          else
+            git commit -m '[auto][on-merge-queue] Update input variations tests'
+            git push
+          fi
+
   model-autogen-op-tests:
-    needs: lowering-tests
+    needs: [push-autogen-op-tests]
     if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.commit_report != 'None'}}
```
ayerofieiev-tt, line 142: consider something like
```
    if: ${{ github.event_name == 'workflow_dispatch' && (github.event.inputs.commit_report == 'Tests' || github.event.inputs.commit_report == 'All')}}
```

PR #566, opened 2024-12-06
Fuse mat mul with activation following it

boris-drazic:
### Problem description
`torch.ops.aten.mm.default` and `torch.ops.aten.bmm.default` are lowered to `ttnn.mm`, which has option to execute an activation function together with the multiplication. The three supported functions are `torch.nn.functional.relu`, `torch.nn.functional.gelu`, and `torch.nn.functional.silu`. Executing activation together with multiplication is faster, so mat mul and activation function should be fused.
     
### What's changed
`MMFusionPass` added to perform the fusion of mat mul and activation function that follows it.
Added `tests/optimization/test_mm_fusion.py` to test fusion.

### DIFF
--- a/tests/lowering/matmul/test_matmul.py
+++ b/tests/lowering/matmul/test_matmul.py
@@ -36,3 +36,4 @@ def test_matmul(device, input_shapes):
 
     # Check inference result
     assert_with_pcc(result_before, result_after, 0.99)
+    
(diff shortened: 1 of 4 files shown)


ayerofieiev-tt:
Please run `pre-commit install` in the repo root.

ayerofieiev-tt (commented):
@boris-drazic please look how fusion is handled here

https://github.com/tenstorrent/pytorch2.0_ttnn/blob/main/torch_ttnn/patterns/linear.py

PR #569, opened 2024-12-08
Try lowering `aten._scaled_dot_product_flash_attention`

### LINKED ISSUE #541: aten._scaled_dot_product_flash_attention.default

jdh8:
### Ticket
- Resolves #541

### Problem description
Convert `aten._scaled_dot_product_flash_attention` to a series of ops.  Future goal might be implementing it as a composite kernel op instead.

The source op is functionally equivalent to its high-level counterpart:
https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html

### What's changed

- Remove fallback for torch.ops.aten._scaled_dot_product_flash_attention.default. 
- since ttnn.transformer.scaled_dot_product_attention requires last dimension padded to tile_size, I added padding nodes before and slicing after if needed. 
- Remove getitem(out, 0) node after ttnn.transformer.scaled_dot_product_attention and fix metadata, since it returns only 1 value (unlike torch self attention)
- add unittest for sdpa
- Add get_add_custom_object_in_graph and get_emplace_custom_object_in_graph function to simplify adding custom objects in fx graph

### DIFF
--- a/tests/lowering/misc/test_scaled_dot_product_attention.py
+++ b/tests/lowering/misc/test_scaled_dot_product_attention.py
@@ -21,6 +21,7 @@ def fa_rand(*shape, dtype=torch.bfloat16):
     bernoulli = torch.bernoulli(torch.full(shape, 0.001, dtype=dtype))
     return normal_1 + normal_2 * bernoulli
 
+
 @pytest.mark.parametrize(
     "input_shape, is_causal",
     (
@@ -35,7 +36,6 @@ def fa_rand(*shape, dtype=torch.bfloat16):
         ((1, 16, 1370, 80), True),
         ((1, 12, 1, 64), True),
         ((1, 12, 4, 64), True),
-
         ((1, 16, 197, 64), False),
         ((1, 12, 197, 64), False),
         ((1, 16, 50, 64), False),
@@ -69,4 +69,4 @@ def test_sdpa(device, input_shape, is_causal):
     nodes = [node.target for node in option._out_fx_graphs[0].nodes]
     assert torch.ops.aten._scaled_dot_product_flash_attention.default not in nodes
     assert nodes.count(ttnn.transformer.scaled_dot_product_attention) == 1
-    assert_with_pcc(result_before, result_after, 0.994) # <- Same pcc as in ttnn tests
+    assert_with_pcc(result_before, result_after, 0.994)  # <- Same pcc as in ttnn tests
--- a/tests/lowering/misc/test_scaled_dot_product_attention.py
+++ b/tests/lowering/misc/test_scaled_dot_product_attention.py
@@ -1,0 +1,50 @@
+import pytest
+import torch
+import torch_ttnn
+import ttnn
+
+from tests.utils import assert_with_pcc
+
+
+class ScaledDotProductAttentionModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, *args, **kwargs):
+        return torch.nn.functional.scaled_dot_product_attention(*args, **kwargs)
+
+
+@pytest.mark.parametrize(
+    "input_shape, is_causal",
+    (
+        ((1, 16, 197, 64), False),
+        ((1, 12, 197, 64), False),
+        ((1, 16, 50, 64), False),
+        ((1, 8, 4096, 40), False),
+        ((1, 8, 1024, 80), False),
+        ((1, 8, 256, 160), False),
+        ((1, 8, 64, 160), False),
+        ((1, 12, 50, 64), False),
+        ((1, 16, 1370, 80), False),
+        ((1, 12, 1, 64), False),
+        ((1, 12, 4, 64), True),
+    ),
+)
+def test_sdpa(device, input_shape, is_causal):
+    module = ScaledDotProductAttentionModule()
+    query = torch.rand(input_shape, dtype=torch.bfloat16)
+    key = torch.rand(input_shape, dtype=torch.bfloat16)
+    value = torch.rand(input_shape, dtype=torch.bfloat16)
+    result_before = module.forward(query, key, value, is_causal=is_causal)
+
+    option = torch_ttnn.TorchTtnnOption(device=device, gen_graphviz=False)
+    # The compilation is lazy, so we need to run forward once to trigger the compilation
+    module = torch.compile(module, backend=torch_ttnn.backend, options=option)
+    result_after = module.forward(query, key, value, is_causal=is_causal)
+    option._out_fx_graphs[0].print_tabular()
+
+    # Check the graph has be rewritten and contain ttnn ops
+    nodes = [node.target for node in option._out_fx_graphs[0].nodes]
+    assert torch.ops.aten._scaled_dot_product_flash_attention.default not in nodes
+    assert nodes.count(ttnn.transformer.scaled_dot_product_attention) == 1
+    assert_with_pcc(result_before, result_after)
--- a/tests/lowering/misc/test_scaled_dot_product_attention.py
+++ b/tests/lowering/misc/test_scaled_dot_product_attention.py
@@ -21,6 +21,18 @@ def rand_in_range(shape, a, b, dtype=torch.bfloat16):
 @pytest.mark.parametrize(
     "input_shape, is_causal",
     (
+        ((1, 16, 197, 64), True),
+        ((1, 12, 197, 64), True),
+        ((1, 16, 50, 64), True),
+        ((1, 8, 4096, 40), True),
+        ((1, 8, 1024, 80), True),
+        ((1, 8, 256, 160), True),
+        ((1, 8, 64, 160), True),
+        ((1, 12, 50, 64), True),
+        ((1, 16, 1370, 80), True),
+        ((1, 12, 1, 64), True),
+        ((1, 12, 4, 64), True),
+
         ((1, 16, 197, 64), False),
         ((1, 12, 197, 64), False),
         ((1, 16, 50, 64), False),
@@ -31,10 +43,11 @@ def rand_in_range(shape, a, b, dtype=torch.bfloat16):
         ((1, 12, 50, 64), False),
         ((1, 16, 1370, 80), False),
         ((1, 12, 1, 64), False),
-        ((1, 12, 4, 64), True),
+        ((1, 12, 4, 64), False),
     ),
 )
 def test_sdpa(device, input_shape, is_causal):
+    torch.manual_seed(0)
     module = ScaledDotProductAttentionModule()
     # Values must be centered around 0 to avoid accuracy issues
     query = rand_in_range(input_shape, -10.0, 10.0, dtype=torch.bfloat16)
@@ -52,4 +65,4 @@ def test_sdpa(device, input_shape, is_causal):
     nodes = [node.target for node in option._out_fx_graphs[0].nodes]
     assert torch.ops.aten._scaled_dot_product_flash_attention.default not in nodes
     assert nodes.count(ttnn.transformer.scaled_dot_product_attention) == 1
-    assert_with_pcc(result_before, result_after, 0.99)
+    assert_with_pcc(result_before, result_after, 0.997)
--- a/tests/lowering/misc/test_scaled_dot_product_attention.py
+++ b/tests/lowering/misc/test_scaled_dot_product_attention.py
@@ -14,6 +14,10 @@ class ScaledDotProductAttentionModule(torch.nn.Module):
         return torch.nn.functional.scaled_dot_product_attention(*args, **kwargs)
 
 
+def rand_in_range(shape, a, b, dtype=torch.bfloat16):
+    rand_tensor = torch.rand(*shape, dtype=dtype)
+    return a + rand_tensor * (b - a)
+
 @pytest.mark.parametrize(
     "input_shape, is_causal",
     (
@@ -32,9 +36,10 @@ class ScaledDotProductAttentionModule(torch.nn.Module):
 )
 def test_sdpa(device, input_shape, is_causal):
     module = ScaledDotProductAttentionModule()
-    query = torch.rand(input_shape, dtype=torch.bfloat16)
-    key = torch.rand(input_shape, dtype=torch.bfloat16)
-    value = torch.rand(input_shape, dtype=torch.bfloat16)
+    # Values must be centered around 0 to avoid accuracy issues
+    query = rand_in_range(input_shape, -10.0, 10.0, dtype=torch.bfloat16)
+    key = rand_in_range(input_shape, -10.0, 10.0, dtype=torch.bfloat16)
+    value = rand_in_range(input_shape, -10.0, 10.0, dtype=torch.bfloat16)
     result_before = module.forward(query, key, value, is_causal=is_causal)
 
     option = torch_ttnn.TorchTtnnOption(device=device, gen_graphviz=False)
@@ -47,4 +52,4 @@ def test_sdpa(device, input_shape, is_causal):
     nodes = [node.target for node in option._out_fx_graphs[0].nodes]
     assert torch.ops.aten._scaled_dot_product_flash_attention.default not in nodes
     assert nodes.count(ttnn.transformer.scaled_dot_product_attention) == 1
-    assert_with_pcc(result_before, result_after)
+    assert_with_pcc(result_before, result_after, 0.99)
--- a/tests/lowering/misc/test_scaled_dot_product_attention.py
+++ b/tests/lowering/misc/test_scaled_dot_product_attention.py
@@ -14,9 +14,12 @@ class ScaledDotProductAttentionModule(torch.nn.Module):
         return torch.nn.functional.scaled_dot_product_attention(*args, **kwargs)
 
 
-def rand_in_range(shape, a, b, dtype=torch.bfloat16):
-    rand_tensor = torch.rand(*shape, dtype=dtype)
-    return a + rand_tensor * (b - a)
+# According to https://arxiv.org/pdf/2407.08608
+def fa_rand(*shape, dtype=torch.bfloat16):
+    normal_1 = torch.randn(shape, dtype=dtype)
+    normal_2 = torch.randn(shape, dtype=dtype) * 10
+    bernoulli = torch.bernoulli(torch.full(shape, 0.001, dtype=dtype))
+    return normal_1 + normal_2 * bernoulli
 
 @pytest.mark.parametrize(
     "input_shape, is_causal",
@@ -49,10 +52,11 @@ def rand_in_range(shape, a, b, dtype=torch.bfloat16):
 def test_sdpa(device, input_shape, is_causal):
     torch.manual_seed(0)
     module = ScaledDotProductAttentionModule()
-    # Values must be centered around 0 to avoid accuracy issues
-    query = rand_in_range(input_shape, -10.0, 10.0, dtype=torch.bfloat16)
-    key = rand_in_range(input_shape, -10.0, 10.0, dtype=torch.bfloat16)
-    value = rand_in_range(input_shape, -10.0, 10.0, dtype=torch.bfloat16)
+
+    query = fa_rand(*input_shape, dtype=torch.bfloat16)
+    key = fa_rand(*input_shape, dtype=torch.bfloat16)
+    value = fa_rand(*input_shape, dtype=torch.bfloat16)
+
     result_before = module.forward(query, key, value, is_causal=is_causal)
 
     option = torch_ttnn.TorchTtnnOption(device=device, gen_graphviz=False)
@@ -65,4 +69,4 @@ def test_sdpa(device, input_shape, is_causal):
     nodes = [node.target for node in option._out_fx_graphs[0].nodes]
     assert torch.ops.aten._scaled_dot_product_flash_attention.default not in nodes
     assert nodes.count(ttnn.transformer.scaled_dot_product_attention) == 1
-    assert_with_pcc(result_before, result_after, 0.997)
+    assert_with_pcc(result_before, result_after, 0.994) # <- Same pcc as in ttnn tests
--- a/torch_ttnn/backend.py
+++ b/torch_ttnn/backend.py
@@ -71,20 +71,6 @@ class TorchTtnnOption:
         self._ttnn_fx_graphs = list()
 
 
-HighAccuracyComputeConfig = ttnn.WormholeComputeKernelConfig(
-    math_fidelity=ttnn.MathFidelity.HiFi4,
-    math_approx_mode=False,
-    fp32_dest_acc_en=False,
-    packer_l1_acc=False,
-)
-
-MidAccuracyComputeConfig = ttnn.WormholeComputeKernelConfig(
-    math_fidelity=ttnn.MathFidelity.HiFi2,
-    math_approx_mode=True,
-    fp32_dest_acc_en=False,
-    packer_l1_acc=False,
-)
-
 def register_ttnn_objects(option: TorchTtnnOption):
     """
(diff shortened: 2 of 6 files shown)


jerrysky3:
ttnn seems to already have the corresponding API https://docs.tenstorrent.com/ttnn/latest/ttnn/api/ttnn.transformer.scaled_dot_product_attention.html, can it be used here?

swimdi:
There's its coressponding test

https://github.com/tenstorrent/tt-metal/blob/main/tests/tt_eager/python_api_testing/unit_testing/misc/test_scaled_dot_product_attention.py

ayerofieiev-tt:
@jdh8 please let us know whatever question you got, I will help to connect with the right stakeholder to resolve this fast.

tests/lowering/misc/test_scaled_dot_product_attention.py
```diff
@@ -0,0 +1,50 @@
...
+        ((1, 12, 197, 64), False),
+        ((1, 16, 50, 64), False),
+        ((1, 8, 4096, 40), False),
+        ((1, 8, 1024, 80), False),
+        ((1, 8, 256, 160), False),
+        ((1, 8, 64, 160), False),
+        ((1, 12, 50, 64), False),
+        ((1, 16, 1370, 80), False),
+        ((1, 12, 1, 64), False),
+        ((1, 12, 4, 64), True),
```
jdh8, line 30: Inferred 0 batch size :thinking:

```
FAILED tests/lowering/misc/test_scaled_dot_product_attention.py::test_sdpa[input_shape0-False] - AssertionError: list(expected_pytorch_result.shape)=[1, 16, 197, 64] vs list(actual_pytorch_result.shape)=[0, 16, 197, 64]
FAILED tests/lowering/misc/test_scaled_dot_product_attention.py::test_sdpa[input_shape1-False] - AssertionError: list(expected_pytorch_result.shape)=[1, 12, 197, 64] vs list(actual_pytorch_result.shape)=[0, 12, 197, 64]
FAILED tests/lowering/misc/test_scaled_dot_product_attention.py::test_sdpa[input_shape2-False] - AssertionError: list(expected_pytorch_result.shape)=[1, 16, 50, 64] vs list(actual_pytorch_result.shape)=[0, 16, 50, 64]
FAILED tests/lowering/misc/test_scaled_dot_product_attention.py::test_sdpa[input_shape3-False] - AssertionError: list(expected_pytorch_result.shape)=[1, 8, 4096, 40] vs list(actual_pytorch_result.shape)=[0, 8, 4096, 40]
FAILED tests/lowering/misc/test_scaled_dot_product_attention.py::test_sdpa[input_shape4-False] - AssertionError: list(expected_pytorch_result.shape)=[1, 8, 1024, 80] vs list(actual_pytorch_result.shape)=[0, 8, 1024, 80]
FAILED tests/lowering/misc/test_scaled_dot_product_attention.py::test_sdpa[input_shape5-False] - AssertionError: list(expected_pytorch_result.shape)=[1, 8, 256, 160] vs list(actual_pytorch_result.shape)=[0, 8, 256, 160]
FAILED tests/lowering/misc/test_scaled_dot_product_attention.py::test_sdpa[input_shape6-False] - AssertionError: list(expected_pytorch_result.shape)=[1, 8, 64, 160] vs list(actual_pytorch_result.shape)=[0, 8, 64, 160]
FAILED tests/lowering/misc/test_scaled_dot_product_attention.py::test_sdpa[input_shape7-False] - AssertionError: list(expected_pytorch_result.shape)=[1, 12, 50, 64] vs list(actual_pytorch_result.shape)=[0, 12, 50, 64]
FAILED tests/lowering/misc/test_scaled_dot_product_attention.py::test_sdpa[input_shape8-False] - AssertionError: list(expected_pytorch_result.shape)=[1, 16, 1370, 80] vs list(actual_pytorch_result.shape)=[0, 16, 1370, 80]
FAILED tests/lowering/misc/test_scaled_dot_product_attention.py::test_sdpa[input_shape9-False] - AssertionError: list(expected_pytorch_result.shape)=[1, 12, 1, 64] vs list(actual_pytorch_result.shape)=[0, 12, 1, 64]
FAILED tests/lowering/misc/test_scaled_dot_product_attention.py::test_sdpa[input_shape10-True] - AssertionError: list(expected_pytorch_result.shape)=[1, 12, 4, 64] vs list(actual_pytorch_result.shape)=[0, 12, 4, 64]
=================================================================== 11 failed in 14.22s ====================================================================
                 Device | INFO     | Closing user mode device drivers
```

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -423,6 +423,21 @@ def is_zero_dim(meta):
...
+                if dropout_p > 0.0:
+                    return self.call_function_prop_meta(target, args, kwargs)
+
+                return self.call_function_prop_meta(
+                    ttnn.transformer.scaled_dot_product_attention,
+                    args[:3],
+                    {"is_causal": is_causal},
+                )
+
+            return select(*args[3:])
```
ayerofieiev-tt, line 439: need to fire a ticket for unsupported cases?

tests/lowering/misc/test_scaled_dot_product_attention.py
ayerofieiev-tt: Is this still an issue?

ayerofieiev-tt (approved):

tests/lowering/misc/test_scaled_dot_product_attention.py
jdh8: I just confirmed that this issue still lingers.  I filed tenstorrent/tt-metal#16021 to keep track on this.

torch_ttnn/passes/lowering/to_tt_pass.py
jdh8: I'm marking this issue as a feature request tenstorrent/tt-metal#16022.  No input variation now has nonzero `dropout_p` yet.  It's still good to keep an eye.

cglagovichTT: Doesn't this logic drop the attention mask, which must be provided if `is_causal == False`

jdh8: [`aten._scaled_dot_product_flash_attention`](https://pytorch.org/cppdocs/api/function_namespaceat_1a99894de13fe7feccc3bcf084ee870f02.html) does not provide attention mask as far as I know.  I have not yet found better documentation.  Please correct me if I am wrong.

tests/lowering/misc/test_scaled_dot_product_attention.py
philei-tt: It seems like the issue was that torch.ops.aten._scaled_dot_product_flash_attention.default returns a tuple of values and torch.fx inserts getitem(ret, 0) after it. I fixed this in the last commit. 

I'm not 100% sure if that's it, but I checked in debug that it returns 9 values, just like `_scaled_dot_product_fused_attention_overrideable` from torch: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L14884, which are: 

(Tensor output, Tensor logsumexp, Tensor cum_seq_q, Tensor cum_seq_k, SymInt max_q, SymInt max_k, Tensor philox_seed, Tensor philox_offset, Tensor debug_attn_mask) 

I think those are probably used for debugging purposes.

ayerofieiev-tt: Outputs can be used for various purposes. 
Most typically to save for a backward pass, but not limited to that

philei-tt:
After fixing integration with torch.fx, two issues appeared: 
Poor pcc. Might be fixed with higher fidelity: https://github.com/tenstorrent/tt-metal/issues/20710 
Fail on shape not divisibe by tile size. https://github.com/tenstorrent/tt-metal/issues/20715 

The latter is blocking this PR

philei-tt:
@ayerofieiev-tt @kevinwuTT I think it's done, please take a look. Both new unittest and all torchvision models using self attention pass. Sometimes accuracy even increased: 

vit_b_16
vit_b_32
vit_l_16
vit_l_32
vit_h_14:
```
>> TTNN flash attention:
2025-04-23 15:39:43,993 - INFO - Accuracy calculated: 0.995213894686952.
2025-04-23 15:39:49,908 - INFO - Accuracy calculated: 0.9882545445126941.
2025-04-23 15:40:01,502 - INFO - Accuracy calculated: 0.9974409892199139.
2025-04-23 15:40:10,571 - INFO - Accuracy calculated: 0.9912243804313401.
2025-04-23 15:40:42,623 - INFO - Accuracy calculated: 0.9868593309834357.

>> fallback torch:
2025-04-23 15:48:03,483 - INFO - Accuracy calculated: 0.9956420948208853.
2025-04-23 15:48:08,900 - INFO - Accuracy calculated: 0.9868819893967631.
2025-04-23 15:48:19,833 - INFO - Accuracy calculated: 0.9970342743148297.
2025-04-23 15:48:28,328 - INFO - Accuracy calculated: 0.9904900120485003.
2025-04-23 15:48:59,469 - INFO - Accuracy calculated: 0.9726238324191475.
```

Please check out new `get_add_custom_object_in_graph` and `get_emplace_custom_object_in_graph` [functions](https://github.com/tenstorrent/pytorch2.0_ttnn/blob/f8563f5e49f8facb0cd570ad12b9249eb92187ab/torch_ttnn/utils.py#L142) 

When trying to add config files to attention, I noticed that registering a new object in graph is very tedious, and you have to create object somewhere else and then use proxy to pass it around. Those function allow to construct custom objects on the fly (directly inside lowering code)

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1295,6 +1305,99 @@ def reshape_1d(code, args=args, kwargs=kwargs):
...
+                    if pad:
+                        output_shape = list(res_node.meta["val"][0].shape)
+                        slice_start = [0] * len(output_shape)
+                        slice_end = output_shape.copy()
+                        slice_end[-1] = d  # Set last dimension back to original size
+                        res_node = g.call_function(ttnn.slice, (res_node, slice_start, slice_end))
+
+                    # torch.ops.aten._scaled_dot_product_flash_attention.default return a tuple of values and inserts a
+                    # getitem(ret, 0) after it. ttnn.transformer.scaled_dot_product_attention only returns one value.
+                    if (val := res_node.meta.get("val", None)) is not None:
```
jmalone-tt, line 1386: Do we need to guard against converting if there are users other than getitem(ret, 0)? Example is how we handle max_pool2d_with_indices

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1295,6 +1305,99 @@ def reshape_1d(code, args=args, kwargs=kwargs):
...
+                    # TODO: Those configs can be further optimized based on input shape to get
+                    # best performance/accuracy tradeoff
+                    compute_kernel_config = get_emplace_custom_object_in_graph(
+                        ttnn.WormholeComputeKernelConfig,
+                        math_fidelity=ttnn.MathFidelity.HiFi2,
+                        math_approx_mode=True,
+                        fp32_dest_acc_en=False,
+                        packer_l1_acc=False,
+                    )
+                    progra_config = get_emplace_custom_object_in_graph(
```
jmalone-tt, line 1357: progra_config -> program_config

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1295,6 +1305,99 @@ def reshape_1d(code, args=args, kwargs=kwargs):
...
+
+                        # 2) rescale Q so that Q*K^T / sqrt(d') == Q_p*K_p^T / sqrt(d)
+                        if scale is None:
+                            calc_scale = math.sqrt((d + pad) / d)
+                            q = g.call_function(ttnn.multiply, args=(q, calc_scale))
+
+                    return q, k, v, d
+
+                def select(dropout_p=0.0, is_causal=False):
+                    # TODO(jdh8): Add suuport for training mode
```
jmalone-tt, line 1338: suuport -> support

torch_ttnn/utils.py
```diff
@@ -131,3 +132,53 @@ def __repr__(self):
...
+    class WrapperObj:
+        def __repr__(self):
+            return key
+
+    torch.fx.graph._register_custom_builtin(key, "", obj)
+    __custom_objects_registry[key] = (WrapperObj(), obj)
+    return __custom_objects_registry[key][0]
+
+
+def get_emplace_custom_object_in_graph(object_type, *args, **kwargs):
```
kevinwuTT, line 155: This looks like an interesting approach. Can you write a small docstring on how this works and how to use this?

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1295,6 +1305,99 @@ def reshape_1d(code, args=args, kwargs=kwargs):
...
+                        ttnn.WormholeComputeKernelConfig,
+                        math_fidelity=ttnn.MathFidelity.HiFi2,
+                        math_approx_mode=True,
+                        fp32_dest_acc_en=False,
+                        packer_l1_acc=False,
+                    )
+                    progra_config = get_emplace_custom_object_in_graph(
+                        ttnn.SDPAProgramConfig,
+                        compute_with_storage_grid_size=device.compute_with_storage_grid_size(),
+                        q_chunk_size=32,
```
jmalone-tt, line 1360: might be worth using a variable for chunk size similar to pad_qkv_ttnn so we don't have these magic numbers. Reminds me I need to do the same for one of my recent changes

torch_ttnn/backend.py
```diff
@@ -72,14 +73,14 @@ def register_ttnn_objects(option: TorchTtnnOption):
...
+    get_add_custom_object_in_graph("ttnn_TILE_LAYOUT", ttnn.TILE_LAYOUT)
 
-    torch.fx.graph._register_custom_builtin("ttnn_uint32", "", ttnn.uint32)
-    torch.fx.graph._register_custom_builtin("ttnn_int32", "", ttnn.int32)
-    torch.fx.graph._register_custom_builtin("ttnn_bfloat16", "", ttnn.bfloat16)
+    get_add_custom_object_in_graph("ttnn_uint32", ttnn.uint32)
+    get_add_custom_object_in_graph("ttnn_int32", ttnn.int32)
+    get_add_custom_object_in_graph("ttnn_bfloat16", ttnn.bfloat16)
 
     torch.fx.graph._register_custom_builtin(
```
kevinwuTT, line 85: Should we change these to use `get_add_custom_object_in_graph` as well?

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1295,6 +1305,99 @@ def reshape_1d(code, args=args, kwargs=kwargs):
...
+                    # Pad last dimension of Q, K, V to tile size
+                    q, k, v = args[:3]
+                    q_padded, k_padded, v_padded, d = pad_qkv_ttnn(
+                        q, k, v, scale=kwargs.get("scale", None), tile_size=32
+                    )
+
+                    # TODO: Those configs can be further optimized based on input shape to get
+                    # best performance/accuracy tradeoff
+                    compute_kernel_config = get_emplace_custom_object_in_graph(
+                        ttnn.WormholeComputeKernelConfig,
```
jmalone-tt, line 1351: Do we need to guard depending on the device (is gray skull / black hole supported here?) Is it just a perf impact or does the ComputeKernelConfig impact correctness?

torch_ttnn/utils.py
```diff
@@ -131,3 +132,53 @@ def __repr__(self):
...
+        repr_str = f"{object_type.__name__}"
+        for arg in args:
+            repr_str += f"_{sanitize(arg)}"
+        for k, v in kwargs.items():
+            repr_str += f"_{sanitize(k)}_{sanitize(v)}"
+        return repr_str
+
+    key = build_repr()
+
+    if key in __custom_objects_registry:
```
jmalone-tt, line 174: Can reuse get_add_custom_object_in_graph for the lower part of this function

torch_ttnn/passes/lowering/to_tt_pass.py
kevinwuTT, line 1357: Possible typo? `program_config`?

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -423,6 +425,14 @@ def _get_output_val(self, node):
             return torch.reshape(self._get_val(args[0]), args[1])
         if node.target == ttnn.permute:
             return torch.permute(self._get_val(args[0]), args[1])
+        if node.target == ttnn.pad:
```
jmalone-tt, line 428: Please add a comment about why we need to convert back to torch for pad /  when we can remove this special case

jmalone-tt (commented):
Thanks for adding comments explaining why certain parts are needed / explaining regex. Left a couple small changes but looks good overall!

kevinwuTT (approved):
Looks good!

The emplace registration of builtins is interesting. I think I can leverage some of that for exporting code as well.

torch_ttnn/passes/lowering/to_tt_pass.py
philei-tt: Okay, will do. The reason is that ttnn padding if output_shape (Shape of output tensor) and input_tensor_start (Start indices to place input tensor in output tensor), while torch's is (pad_start, pad_end, .. for all dimensions) https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html

torch_ttnn/passes/lowering/to_tt_pass.py
philei-tt: Seems like fx always puts getitem(res, out) after flash attention, at least I didn't see other users in my experiments. That might be different for backward pass though

torch_ttnn/passes/lowering/to_tt_pass.py
philei-tt: Great question. grayskull is deprecated and  BlackholeComputeKernelConfig = WormholeComputeKernelConfig: [ttnn/ttnn/types.py](https://github.com/tenstorrent/tt-metal/blob/e2a722a51c3a0aea2f80f9bf1e9f4e57191337e7/ttnn/ttnn/types.py#L51), but I can try adding device specialization.

philei-tt: Do we have some single location (file) with constant values to put arch names there?

torch_ttnn/utils.py
philei-tt: Yeah good catch

philei-tt: I was going to do that but forgot lol

torch_ttnn/passes/lowering/to_tt_pass.py
jmalone-tt: Ok, if it won't break this PR, I would be in favor of adding the guard then. That way we guarantee our assumptions

torch_ttnn/passes/lowering/to_tt_pass.py
jmalone-tt: Looks like we don't have many places that guard, but in `to_tt_pass.py` we have the following line:
58:    return ttnn.device.is_grayskull(device) and target in ops_incompatible_with_grayskull

I don't know if we're ok to just remove all checking against grayskull - might be worth discussing with the team. Guessing there's a similar function to check if device is blackhole. We have ttnn.device.is_wormhole_b0(device) for wormhole

torch_ttnn/passes/lowering/to_tt_pass.py
ayerofieiev-tt: 😢

ayerofieiev-tt: @philei-tt , can you raise this with @ntarafdar please?

philei-tt: Okay

philei-tt: Turns out we can store padding in List[Tuple[int, int]], still not the same, but closer to pytorch's interface. I will still open an issue

torch_ttnn/backend.py
philei-tt: Yes, missed that

torch_ttnn/utils.py
philei-tt: Good idea, will do

philei-tt: Added

philei-tt:
@ayerofieiev-tt @joelsmithTT @kevinwuTT I addressed the comments, please take a look

ayerofieiev-tt (approved):

kevinwuTT:
Looks good. Thanks! Although I think you were meant to tag @jmalone-tt 😅

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1298,6 +1304,109 @@ def reshape_1d(code, args=args, kwargs=kwargs):
...
+                    if pad:
+                        output_shape = list(res_node.meta["val"][0].shape)
+                        slice_start = [0] * len(output_shape)
+                        slice_end = output_shape.copy()
+                        slice_end[-1] = d  # Set last dimension back to original size
+                        res_node = g.call_function(ttnn.slice, (res_node, slice_start, slice_end))
+
+                    # torch.ops.aten._scaled_dot_product_flash_attention.default return a tuple of values and inserts a
+                    # getitem(ret, 0) after it. ttnn.transformer.scaled_dot_product_attention only returns one value.
+                    if (val := res_node.meta.get("val", None)) is not None and is_getitem_0_only_user(node):
```
jmalone-tt, line 1395: I apologize, I must not have explained the comment about getitem_0_only_user very well. We should check `is_getitem_0_only_user` to guard against lowering `torch.ops.aten._scaled_dot_product_flash_attention.default` to `ttnn.transformer.scaled_dot_product_attention` at all. For example, if we try to lower `max_pool2d_with_indices` but the indices are actually used, the function returns None, which corresponds to not lowering the function

torch_ttnn/passes/lowering/to_tt_pass.py
jmalone-tt: I'm still not understanding here. Even if the arguments are formatted differently (which isn't great), can't we keep the ttnn.pad version? Or is there something the torch pad can do that the ttnn pad cannot?

jmalone-tt (changes requested):

torch_ttnn/passes/lowering/to_tt_pass.py
philei-tt: We use ttnn, this code snippet just calculates "val" metadata which is an example of output tensor of the operation. You can see there are other ops that are lowered, like ttnn.add, mul, sub, etc.

torch_ttnn/passes/lowering/to_tt_pass.py
philei-tt: Gotcha, updated

torch_ttnn/passes/lowering/to_tt_pass.py
jmalone-tt: Thanks for clarifying, I thought this was in another part of the file. Change here makes sense then!

jmalone-tt (approved):
Thanks for making those last couple updates! LGTM

torch_ttnn/passes/lowering/to_tt_pass.py
philei-tt: @ayerofieiev-tt https://github.com/tenstorrent/tt-metal/issues/21198

PR #604, opened 2024-12-14
Fix some `aten.add/mul` not lowered

swimdi:
### Ticket
N/A

### Problem description
Some ✅ modes become 🚧 
 - BERT
 - SqueezeBERT
 - Perceiver IO
 - albert/albert-base-v2-classification
 - twmkn9/albert-base-v2-squad2

And the reason is `aten.add/mul` become not lowered

I think the root cause is because `get_shape(node)` not handle the scalar case

### What's changed
 - Let `get_shape(node)` handle the arg is scalar case, and above model is back to ✅

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -457,9 +457,12 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool
                 return g.call_function(target_wrappers.pack_to_tuple, (output,))
 
             def lower_binary_eltwise(fn, args):
+                if any(isinstance(x, (int, float)) for x in args):
+                    return g.call_function(fn, args)
+
                 shapes = get_shape(args[0]), get_shape(args[1])
 
-                if any(s is None for s in shapes):
+                if any(not s for s in shapes):
                     return None
 
                 if max(map(len, shapes)) > 4 and shapes[0][-3:-2] != shapes[1][-3:-2]:
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -459,6 +459,12 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool
             def lower_binary_eltwise(fn, args):
                 shapes = get_shape(args[0]), get_shape(args[1])
 
+                if (isinstance(args[0], torch.fx.node.Node) and shapes[0] == torch.Size()) or (
+                    isinstance(args[1], torch.fx.node.Node) and shapes[1] == torch.Size()
+                ):
+                    # ttnn.from_torch not yet support scalar tensor, see issue 442
+                    return None
+
                 if any(s is None for s in shapes):
                     return None
 
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -457,12 +457,9 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool
                 return g.call_function(target_wrappers.pack_to_tuple, (output,))
 
             def lower_binary_eltwise(fn, args):
-                if any(isinstance(x, (int, float)) for x in args):
-                    return g.call_function(fn, args)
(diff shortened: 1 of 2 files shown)


ayerofieiev-tt (approved):

jdh8 (approved):

jdh8:
I ran the tests in `tests/autogen_op/resnet`.  It seems that `aten.convolution.default` is affected.  I have no idea about how to debug though.

swimdi:
I reinstalled tt-metal yesterday, and it seems to have worked.

jdh8:
The op tests pass, but the model tests still fail.
```
(python_env) jdh8@tt-loudbox:~/pytorch2.0_ttnn$ pytest tests/models/resnet*
=================================================================== test session starts ====================================================================
platform linux -- Python 3.8.10, pytest-7.2.2, pluggy-1.5.0
rootdir: /home/jdh8/pytorch2.0_ttnn/tests, configfile: pytest.ini
plugins: timeout-2.2.0, split-0.8.2, dash-2.15.0, xdist-3.6.1, anyio-4.5.2
collected 4 items                                                                                                                                          

tests/models/resnet/test_resnet.py .                 Always | FATAL    | Index is out of bounds for the rank, should be between 0 and -1 however is 0
                 Always | FATAL    | Index is out of bounds for the rank, should be between 0 and -1 however is 0
E.                                                                                                               [ 50%]
tests/models/resnet50/test_resnet50.py .                 Always | FATAL    | Index is out of bounds for the rank, should be between 0 and -1 however is 0
                 Always | FATAL    | Index is out of bounds for the rank, should be between 0 and -1 however is 0
E.
```

ayerofieiev-tt:
I made a fix to the main of tt-metal and just kick started the release wheel build.
Once that wheel is picked up overnight, please re-run checks during your Monday. You might see things working

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -457,6 +457,9 @@ def batch_norm_inference(input, weight, bias, mean, var, momentum, eps):
                 return g.call_function(target_wrappers.pack_to_tuple, (output,))
 
             def lower_binary_eltwise(fn, args):
+                if any(isinstance(x, (int, float)) for x in args):
```
swimdi: This will allow args[0] with rank > 4 and args[1] is scalar

swimdi:
For bert its argument is -3.3895313892515355e+38 and this lowering pass
```
ttnn_multiply = ttnn_decorators_ttnn_multiply(ttnn_rsub, -3.3895313892515355e+38)
```
For resnet its argument is tensor(1), and this lowering
```
ttnn_from_torch_2 = ttnn_decorators_ttnn_from_torch(primals_65, layout = ttnn_TILE_LAYOUT, device = ttnn_Specified_Device, dtype = ttnn_bfloat16)
ttnn_add_1 = ttnn_decorators_ttnn_add(ttnn_from_torch_2, 1)
```
will cause above err msg on ttnn_from_torch_2

Add this code can pass
```
if (isinstance(args[0], torch.fx.node.Node) and shapes[0] == torch.Size()) or \
   (isinstance(args[1], torch.fx.node.Node) and shapes[1] == torch.Size()):
    return None
```
but if the tt-metal update is to fix the above err msg, then can just wait the tt-metal updated

swimdi:
> For bert its argument is `-3.3895313892515355e+38` and this lowering pass
> ```
> ttnn_multiply = ttnn_decorators_ttnn_multiply(ttnn_rsub, -3.3895313892515355e+38)
> ```
> For resnet its argument `primals_65` is `tensor(1)`, and this lowering
> ```
> ttnn_from_torch_2 = ttnn_decorators_ttnn_from_torch(primals_65, layout = ttnn_TILE_LAYOUT, device = ttnn_Specified_Device, dtype = ttnn_bfloat16)
> ttnn_add_1 = ttnn_decorators_ttnn_add(ttnn_from_torch_2, 1)
> ```
> will cause above err msg on ttnn_from_torch_2
> 
> Add this code can pass
> ```
> if (isinstance(args[0], torch.fx.node.Node) and shapes[0] == torch.Size()) or \
>    (isinstance(args[1], torch.fx.node.Node) and shapes[1] == torch.Size()):
>     return None
> ```
> but if the tt-metal update is to fix the above err msg, then can just wait the tt-metal updated

Seems the curreent update not yet solved this problem, and I think this is #442 problem, so I add the code check if is scalar tensor

Issue #610, opened 2024-12-16
Constant folding need meta["val"]


swimdi:
Take `segformer` for example, one of its `mul` input is
```
(Pdb) print(args)
(_unsafe_index_tensor_1, _folded_sub_2)
(Pdb) get_shape(args[0]), get_shape(args[1])
(torch.Size([1, 256, 128, 128]), None)
```
And because its `args[1]` is constant folded and `meta["val"]` is discarded, so this mul cannot lowered

https://github.com/tenstorrent/pytorch2.0_ttnn/blob/8279be8d4fac07e63f089e2d0c5201cdbef5d5e1/torch_ttnn/passes/lowering/to_tt_pass.py#L468-L469

PR #613, opened 2024-12-17
Get correct tensor shape of `get_attr` op

pschang-phy:
### Ticket
https://github.com/tenstorrent/pytorch2.0_ttnn/issues/610

### Problem description
`get_shape` function could not properly deal with `get_attr` op

### What's changed
Get correct tensor shape of `get_attr` op.

### DIFF
--- a/torch_ttnn/utils.py
+++ b/torch_ttnn/utils.py
@@ -10,6 +10,16 @@ def GraphCleanup(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
 
 
 def get_shape(gm: torch.fx.GraphModule, node_or_shape):
+    """
+    Get the shape of a node or shape itself.
+
+    Args:
+        gm (torch.fx.GraphModule): The GraphModule containing the node.
+        node_or_shape: The node or shape to get the shape of. Can be an int, float, torch.Size, list, tuple, or torch.fx.node.Node.
+
+    Returns:
+        torch.Size or None: The shape of the node or shape itself, or None if it cannot be determined.
+    """
     if isinstance(node_or_shape, (int, float)):
         return torch.Size()
     if isinstance(node_or_shape, (torch.Size, list, tuple)):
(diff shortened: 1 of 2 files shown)


swimdi (approved):

torch_ttnn/utils.py
```diff
@@ -9,14 +9,19 @@ def GraphCleanup(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
...
     if isinstance(node_or_shape, (int, float)):
         return torch.Size()
     if isinstance(node_or_shape, (torch.Size, list, tuple)):
         return node_or_shape
     if isinstance(node_or_shape, torch.fx.node.Node):
         if (val := node_or_shape.meta.get("val", None)) is not None:
             return val.size()
+
+        if node_or_shape.op == "get_attr":
+            val = getattr(gm, node_or_shape.target)
```
swimdi, line 22: Is it possible `val` be type of `int/float` ?

torch_ttnn/utils.py
```diff
@@ -9,14 +9,19 @@ def GraphCleanup(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
...
+            return val.size()
```
swimdi, line 23: Is it possible val be type of int/float ?

swimdi (approved):

torch_ttnn/utils.py
```diff
@@ -9,14 +9,19 @@ def GraphCleanup(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
     return gm
 
 
-def get_shape(node_or_shape):
+def get_shape(gm: torch.fx.GraphModule, node_or_shape):
```
ayerofieiev-tt, line 12: I think this function deserves for some doc string

ayerofieiev-tt (approved):

torch_ttnn/utils.py
pschang-phy: I check the doc, `getattr` could be any type. I add more check for `val` type.

torch_ttnn/utils.py
pschang-phy: agree. I have updated it for doc string b280d540a5e5c31bde8e75d0f312c1b2815419f8

PR #559, opened 2024-12-05
Support lowering of index

swimdi:
### Ticket
#535 

### Problem description
for example, if `input.shape = (3, 4, 5), indices = [tensor([[0, 1, 1]]), tensor([[2, 1, 2]])]`
then its output is `[[input[0][2], input[1][1], input[1][2]]]`
So lowering `aten.index` to several `getitem/reshape/cat`

### What's changed
 - Add `DigestAtenOps` and it digest aten.index to serveral torch ops

I'll merge #532 first then this PR

PR #616, opened 2024-12-17
Decompose `aten.full_like` into `aten.full`

jerrysky3:
### Ticket
#589

### Problem description
Decompose `aten.full_like` into `aten.full`. There is one remaining dynamic shape case is not supported yet and guarded

### What's changed
- Decompose `aten.full_like` into `aten.full`
- Add tests


ayerofieiev-tt (approved):

jerrysky3:
@swimdi fyi I also added a new pass to decompose aten ops into aten ops in this change. It might duplicate with the `DigestAtenOps` in #559

PR #621, opened 2024-12-17
Refine the logic to lowering `aten._to_copy.default`

pschang-phy:
### Ticket
N/A

### Problem description
Some `aten._to_copy.default` remains after compiling.

### What's changed
Refine the lowering logic as below
1. keep it if the graph output uses this op
2. if `dst_dtype` is bool, then transform to equivalent calculation **input != 0**
3. Try cast to supported ttnn type, or keep it if the cast is not supported
4. remove this op


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -825,28 +825,31 @@ def lower_binary_eltwise(fn, args):
...
-                src_dtype = node.args[0].meta["val"].dtype
-                dst_dtype = kwargs["dtype"]
-                # Some aten op need it to cast specific dtype (ex, index_select)
-                # Keep it if casting from int to float or reverse
+
+                src_dtype, dst_dtype = node.args[0].meta["val"].dtype, kwargs["dtype"]
+
+                # casting to bool type is equivalent to checking if the value is not 0
+                if dst_dtype == torch.bool:
+                    return g.call_function(ttnn.ne, args=(args[0], 0))
```
swimdi, line 837: Does the output dtype of `ttnn.ne` bool?

swimdi (approved):

torch_ttnn/passes/lowering/to_tt_pass.py
ayerofieiev-tt: I think it should be, but don’t think it is actually

pschang-phy: The result type of `ttnn.ne` is not bool, it's BF16. But I think it could not need to be.
1. If this bool `to_copy` is used by TTNN op, it all use BF16 computation in TTNN. All we need is to ensure the correct numerical values after casting to bool.
2. If it's return or fallback to torch, the `AddDataMovePass` has handled the correct type after inserting `ttnn.to_torch`
```
%ttnn_to_torch : [num_users=1] = call_function[target=ttnn.decorators.ttnn.to_torch](args = (%ttnn_ne,), kwargs = {dtype: torch.bool})
```

PR #623, opened 2024-12-18
Lowering `aten.stack.default` op

pschang-phy:
### Ticket
https://github.com/tenstorrent/pytorch2.0_ttnn/issues/596

### Problem description
`aten.stack.default` not lowering yet.

### What's changed
`aten.stack.default` op is transform to equivalent computation
1. reshape all tensors with specified dimension
2. concat all unsqueezed tensors


tests/lowering/tensor_manipulation/test_stack.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def forward(self, tensors, dim):
+        return torch.stack(tensors, dim)
+
+
+@pytest.mark.parametrize(
+    "input_shape, dim",
+    [
+        ((12, 16), -1),
+    ],
```
ayerofieiev-tt, line 19: can we get a couple more corner cases captured here?

ayerofieiev-tt (approved):

tests/lowering/tensor_manipulation/test_stack.py
pschang-phy: ok, I will add more different cases

PR #628, opened 2024-12-19
Enable ETH dispatch by default and add logging

ayerofieiev-tt:
### Ticket
None

### Problem description
Some conv cases pass in our sweep tests and fail during a run through a compiler.
One of big differences is that we run on a fewer cores when tests are running on n300.

### What's changed
Enabling ETH dispatch so that more of a device core grid could be used.
Slightly tweak logging


kevinwuTT (approved):

ayerofieiev-tt:
@jerrysky3 , FYI

PR #652, opened 2024-12-19
Let conv2d_transpose rely on automatic sharding

ayerofieiev-tt:
### Ticket
None

### Problem description
`conv2d_transpose` currently provides ConfConfig which prevents the op from automatically choosing sharding.

### What's changed
Remove the conf config


ayerofieiev-tt:
Passed CI tests. Force merging

PR #653, opened 2024-12-19
Add option to compile and run multiple iterations

kevinwuTT:
### Ticket
None

### Problem description
The first iteration of the compiled model always include warm up time. We want to run the model multiple times and report the last iteration.

### What's changed
* Add option to run multiple iterations
* Replace the run time report of the docs with the 5th iteration, but keep a record of the 1st iteration in the metrics file for future use
* Modify the CI to run the multiple iterations only when refreshing the README and docs


tools/collect_metrics.py
```diff
@@ -36,8 +36,8 @@
         "Execution time (in seconds) of the model before conversion.",
     ),
     "compiled_run_time": (
-        "Compiled Run Time (ms)",
-        "Execution time (in seconds) of the model after conversion.",
+        "Compiled Run Time for 5th Iteration (ms)",
```
ayerofieiev-tt, line 39: Lets keep column name short, but be clear in the description on this detail

ayerofieiev-tt (approved):

PR #667, opened 2024-12-23
Move lowered with aten to decompose as possible

swimdi:
### Ticket
N/A

### Problem description
`torch.ops.aten.select.int` and `torch.ops.aten.sum.default` will generated aten op in lowered function, move it to `decompose_aten_to_aten_ops` as possible. `torch.ops.aten.sum.default` cannot fully moved to, so add `again` to check

`aten.select.int` and `aten.sum.default` will generate ATen operations in the lowered function. The former can moved to `decompose_aten_to_aten_ops` but later cannot, so add an `again` mechanism

### What's changed
Edit `torch.ops.aten.select.int` and `torch.ops.aten.sum.default`

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -361,6 +361,7 @@ class GraphWrapper:
 
 
 def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool) -> torch.fx.GraphModule:
+    again = False
     nodes = list(gm.graph.nodes)
     for node in nodes:
         if not can_lowering_to_ttnn(node):
@@ -368,6 +369,7 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool
         g = GraphWrapper(node)
 
         def rewrite_node(node):
+            nonlocal again
             args = node.args
             kwargs = node.kwargs
 
@@ -555,17 +557,6 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool
                 tensor = g.call_function(ttnn.embedding, (args[1], args[0]), {"layout": layout})
                 return tensor if tiled else g.call_function(ttnn.to_layout, (tensor, TtnnTileLayout()))
 
-            if node.target == torch.ops.aten._log_softmax.default:
-                softmax_node = g.call_function(
-                    ttnn.softmax,
-                    args[:2],
-                    {
-                        "numeric_stable": True,
-                        **kwargs,
-                    },
-                )
-                return g.call_function(ttnn.log, (softmax_node,), kwargs)
-
             if node.target == torch.ops.aten.div.Tensor:
                 if isinstance(args[1], (float, int)):
                     return g.call_function(ttnn.mul, (args[0], 1.0 / args[1]), {})
@@ -655,6 +646,9 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool
                 return None
 
             if node.target == torch.ops.aten.squeeze.dim or node.target == torch.ops.aten.squeeze.default:
+                if get_shape(gm, args[0]) in [torch.Size([1]), torch.Size([])]:
+                    # see #442
+                    return None
                 if use_less_ttnn_op_types or node.target == torch.ops.aten.squeeze.default:
                     # ttnn.squeeze does not support calling the OP without provided dim (torch.ops.aten.squeeze.default)
                     # squeezing is the same as reshaping to shape of output tensor of squeeze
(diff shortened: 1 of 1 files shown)


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -361,13 +361,15 @@ def inserting_before(self, node):
...
 def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool) -> torch.fx.GraphModule:
+    again = False
     nodes = list(gm.graph.nodes)
     for node in nodes:
         if not can_lowering_to_ttnn(node):
             continue
         g = GraphWrapper(node)
 
         def rewrite_node(node):
+            nonlocal again
```
ayerofieiev-tt, line 372: Can you clarify what you achieve with this?

swimdi: the `aten.sum` will be lowered to `ttnn.sum` and `aten.squeeze`, so it need to run again to lower `aten.squeeze`. Currently rewrite_node's return is return node, but if let it return (node, again) it will change many code, so I use `nonlocal again` to set it

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1187,6 +1171,20 @@ def decompose_aten_to_aten_ops(g: GraphWrapper, node):
...
+        output_shape = get_shape(gm, node)
+        if input_shape.numel() == output_shape.numel() and len(output_shape) == 0:
+            return g.call_function(torch.ops.aten.squeeze.dim, args=(args[0], 0))
+        return None
+
+    if node.target == torch.ops.aten.sum.default:
+        input_shape = get_shape(gm, args[0])
+        output_shape = get_shape(gm, node)
+        if input_shape.numel() == output_shape.numel():
+            return g.call_function(torch.ops.aten.squeeze.default, args=(args[0],))
```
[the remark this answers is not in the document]
swimdi: This code is from @pschang-phy 
I think if input size equal to output size on `aten.sum`, then we can deduce tha input shape have at most one dimension that is not 1, for example
```
input_shape = [1, 7, 1, 1]
output_shape = [7]
```
So in this case, it equivalence with squeeze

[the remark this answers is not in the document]
jerrysky3: Checked the `torch.sum`, I'm wondering if checking `input_shape.numel() == 1` will be clearer and still correct? @pschang-phy

 But we can revise this later

torch_ttnn/passes/lowering/to_tt_pass.py
jerrysky3: I'm thinking if a simpler way to do this is "if rewrite_node() returns a new node -> graph_changed = True -> run ReplaceMoreTtManually again". WDYT?

This is less efficient (basically we always need to iterate twice) but I assume currrently the graph rewrite doesn't take too much time? To be more efficient, we can later track the new nodes created with `GraphWrapper.call_function` and only scan those nodes in the following iteration

torch_ttnn/passes/lowering/to_tt_pass.py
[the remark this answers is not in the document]
pschang-phy: @jerrysky3 is right. `aten.sum` will sum over all element of input tensor and output a rank-0 tensor (single element).

My original idea is to handle the case of input has only one element (shape=[1] or [1, 1],...), such cases could just do squeeze to rank-0 tensor.

And yes, use the checking `input_shape.numel() == 1` would be more clear.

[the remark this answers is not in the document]
swimdi: > I think if input size equal to output size on aten.sum, then we can deduce tha input shape have at most one dimension that is not 1, for example

oh I give the wrong describe, only `input_shape.numel() == 1` in `input_shape.numel() == output_shape.numel()`

torch_ttnn/passes/lowering/to_tt_pass.py
swimdi: > I'm thinking if a simpler way to do this is "if rewrite_node() returns a new node -> graph_changed = True -> run ReplaceMoreTtManually again". WDYT?

ok, I change to use `modified`

> To be more efficient, we can later track the new nodes created with GraphWrapper.call_function and only scan those nodes in the following iteration

I'm not sure how's a proper way to know it is created with `GraphWrapper.call_function`

swimdi: And once applied `modified`, I think we can cancel the `decompose_aten_to_aten_ops`

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1222,6 +1219,9 @@ def call(self, gm: torch.fx.GraphModule):
         gm = ReplaceMoreTt(gm, self.device, self.use_less_ttnn_op_types).transform()
 
         # Replace patterns manually
-        gm = ReplaceMoreTtManually(gm, self.use_less_ttnn_op_types)
+        while True:
```
jerrysky3, line 1222: Maybe set a threshold and throw an exception if exceeds the limit? So the compiler won't silently run infinite loop

jerrysky3 (approved):

PR #670, opened 2024-12-23
Lower `aten.as_strided` when it can be simplified into `ttnn.reshape`

### LINKED ISSUE #587: aten.as_strided.default

jdh8:
### Ticket
- Partially resolves #587

### Problem description
As per discussion in #587, we only convert `aten.as_strided` that is equivalent to `ttnn.reshape`.  This would convert all our known input variations for now.

> A general implementation is best done as a kernel op. Its element rearrangement is far more complicated than slicing. https://pytorch.org/docs/stable/generated/torch.as_strided.html
> 
> However, our use cases can be simplified into `reshape`. The `stride` argument matches the default strides of `size` for each case. I am working on a conversion that only lowers these cases. https://github.com/tenstorrent/pytorch2.0_ttnn/blob/main/docs/operations/aten.as_strided.default.md

### What's changed
- [x] Convert `aten.as_strided` to `ttnn.reshape` when such simplification is correct
- [x] Pass auto generated test cases
- [x] Pass manual test cases

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -326,6 +326,22 @@ class ReplaceMoreTt(torch.fx.Transformer):
         if target == torch.ops.aten.detach.default:
             return args[0]
 
+        if target == torch.ops.aten.as_strided.default:
+            unpack = lambda tensor, size, stride, offset=0: (tensor, size, stride, offset)
+            tensor, size, stride, offset = unpack(*args)
+            ndims = len(size)
+
+            if offset != 0 or len(stride) != ndims:
+                return self.call_function_prop_meta(target, args, kwargs)
+
+            dummy = torch.empty(0, *size, dtype=torch.bool)
+
+            for i in range(ndims):
+                if stride[i] != dummy.stride(i + 1) and size[i] != 1:
+                    return self.call_function_prop_meta(target, args, kwargs)
+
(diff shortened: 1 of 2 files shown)


jdh8:
CI tests running at https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/12471891387

swimdi (approved):

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -326,6 +326,22 @@ def __init__(self, target, args, kwargs):
         if target == torch.ops.aten.detach.default:
             return args[0]
 
+        if target == torch.ops.aten.as_strided.default:
+            unpack = lambda tensor, size, stride, offset=0: (tensor, size, stride, offset)
+            tensor, size, stride, offset = unpack(*args)
+            ndims = len(size)
+
+            if offset != 0 or len(stride) != ndims:
+                return self.call_function_prop_meta(target, args, kwargs)
```
ayerofieiev-tt, line 335: Why do we do `return None` in some other cases?

ayerofieiev-tt (approved):

torch_ttnn/passes/lowering/to_tt_pass.py
jdh8: I make returning `None` mean skipping conversion in `rewrite_node` nested in `ReplaceMoreTtManually`.  We are in `ReplaceMoreTt.call_function` here, so we skip conversion this way.

Issue #544, opened 2024-12-02
aten._unsafe_index.Tensor


jdh8:

swimdi:
Hi @ayerofieiev-tt , about the question of **what the different between`aten.index.Tensor` and `aten._unsafe_index.Tensor`?**, the answer is they are almost the same
```
input = torch.arange(2*3*4).reshape([2,3,4])
indices = [None, torch.tensor([1, 0]), torch.tensor([2, 3])]
index = torch.ops.aten.index.Tensor(input, indices)
unsafe_index = torch.ops.aten._unsafe_index.Tensor(input, indices)
print((index == unsafe_index).all())

tensor(True)
```
And according to the [source code](https://github.com/pytorch/pytorch/blob/539286a67b117162ecca9e59fb457a59f9504bbd/aten/src/ATen/native/TensorAdvancedIndexing.cpp#L646-L657), the only different is `aten._unsafe_index.Tensor` NOT allow indices dtype be boolean but `aten.index.Tensor`
```
input2 = torch.arange(2).reshape([2])
indices2 = [torch.tensor([True, False])]
index = torch.ops.aten.index.Tensor(input2, indices2) # no problem
unsafe_index = torch.ops.aten._unsafe_index.Tensor(input2, indices2)

RuntimeError: _unsafe_index found unexpected index type Bool
```

So I think these two op can use the same lowering code

Issue #535, opened 2024-12-02
aten.index.Tensor


jdh8:

swimdi:
In #559 it lowering `aten.index.Tensor` to serveral `aten.reshape/getitem/aten.cat`, but `aten.reshape` have no lowering yet

I try to add lowering of `aten.reshape`, but got this error
```
pytest tests/lowering/misc/test_index.py

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
    getitem_3 = ttnn_to_device_1[1]

peError: object of type 'ttnn._ttnn.tensor.Tensor' has no len()
```
It seems failed at `getitem_3 = ttnn_to_device_1[1]`, I think there have two way to solved it

1. find why `getitem_3 = ttnn_to_device_1[1]` failed and TT add the support
2. not use getitem, use `aten.select` and later lowering it to `ttnn.reshape & ttnnn.slice`, but because the `dim, start` is node not int/float, so `aten.select` need to exapand its support

c.c. @ayerofieiev-tt @pschang-phy

PR #682, opened 2024-12-26
Convert some index case to embedding

swimdi:
### Ticket
#535 
#544 
#639 

### Problem description
Some input variation of index like below can be replaced as embedding

https://github.com/tenstorrent/pytorch2.0_ttnn/blob/217470993d314925a6473fe3511081d03d7f1ef8/tests/autogen_op/ALL/test_ALL_aten_index_Tensor.py#L45-L59

So add the conversion of it

### What's changed
 - Convert some case from index to embedding
 - Convert `aten.index_select.Tensor` to `aten.index.Tensor`
 - Add `aten.index_select.Tensor` and `aten.index.Tensor` to const fold pass

credit: @jerrysky3

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -1251,6 +1251,7 @@ def decompose_aten_to_aten_ops(gm: torch.fx.GraphModule, g: GraphWrapper, node):
             index_shape = get_shape(gm, indices[0])
             # magic number, 38000 can pass, 39000 can pass, but 38809 will hang
             # and if device is just ttnn.open_device(device=0), then it can pass
+            # see issue #685
             if index_shape == torch.Size([38809]):
                 return None
             return g.call_function(torch.ops.aten.embedding.default, args=(args[0], indices[0]))
@@ -1294,11 +1295,13 @@ class ToTtPass(PassBase):
         # Decompose some aten ops to simpler aten ops
         max_try = 10
         cnt = 0
-        while cnt < max_try:
+        while True:
             cnt += 1
             gm, modified = rewrite_graph(gm, decompose_aten_to_aten_ops)
             if not modified:
                 break
+            if cnt == max_try:
+                raise RuntimeError("Failed to decompose aten ops to simpler aten ops")
 
         # Replace more patterns with torch.fx.Transformer
         gm = ReplaceMoreTt(gm, self.device, self.use_less_ttnn_op_types).transform()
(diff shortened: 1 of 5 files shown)


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1244,26 +1244,45 @@ def decompose_aten_to_aten_ops(gm: torch.fx.GraphModule, g: GraphWrapper, node):
             return g.call_function(torch.ops.aten.squeeze.default, args=(args[0],))
         return None
 
+    if node.target in [torch.ops.aten.index.Tensor, torch.ops.aten._unsafe_index.Tensor]:
+        input_shape = get_shape(gm, args[0])
+        indices = get_arg(node, 1, "indices")
+        if len(input_shape) == 2 and len(indices) == 1 and indices[0] is not None:
+            index_shape = get_shape(gm, indices[0])
+            # magic number, 38000 can pass, 39000 can pass, but 38809 will hang
```
jerrysky3, line 1252: Maybe file a bug?

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1273,7 +1292,13 @@ def __init__(self, device, use_less_ttnn_op_types):
 
     def call(self, gm: torch.fx.GraphModule):
         # Decompose some aten ops to simpler aten ops
-        gm = rewrite_graph(gm, decompose_aten_to_aten_ops)
+        max_try = 10
+        cnt = 0
+        while cnt < max_try:
```
jerrysky3, line 1297: I'm thinking if we should raise an exception when running out of tries. Because that means the graph still can't not converge and probably in a bad state

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1244,26 +1244,45 @@ def decompose_aten_to_aten_ops(gm: torch.fx.GraphModule, g: GraphWrapper, node):
...
+        if len(input_shape) == 2 and len(indices) == 1 and indices[0] is not None:
+            index_shape = get_shape(gm, indices[0])
+            # magic number, 38000 can pass, 39000 can pass, but 38809 will hang
+            # and if device is just ttnn.open_device(device=0), then it can pass
+            if index_shape == torch.Size([38809]):
+                return None
+            return g.call_function(torch.ops.aten.embedding.default, args=(args[0], indices[0]))
+        return None
+
+    if node.target == torch.ops.aten.index_select.default:
```
jerrysky3, line 1259: Do we have unit tests for index_select and unsafe_index?

jerrysky3 (approved):

torch_ttnn/passes/lowering/to_tt_pass.py
swimdi: ok, I've open an issue

torch_ttnn/passes/lowering/to_tt_pass.py
swimdi: ok, I've add

torch_ttnn/passes/lowering/to_tt_pass.py
swimdi: ok I added

PR #740, opened 2025-02-04
#0: Added a naive batching mechanism for some models in the tests folder

jbedichekTT:
Added a batching flag for the some models in the test folder, repeats the same input tensor across a new dimension.  Also misc supporting functions for that.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -36,7 +36,7 @@ def pytest_addoption(parser):
         help="Run up to the specified iteration count and report metrics based on this iteration.",
     )
     parser.addoption("--gen_op_accuracy_tests", action="store_true")
-
+    parser.addoption("--batch_size", action="store", default=None, help="Batch size for testing")
 
 @pytest.fixture(scope="session")
 def input_var_only_native(request):
@@ -69,6 +69,9 @@ def device():
     ttnn.synchronize_device(device)
     ttnn.close_device(device)
 
+@pytest.fixture(scope="session")
+def get_batch_size(request):
+    return request.config.getoption("--batch_size")
 
 def get_dispatch_core_type():
     # Instead of conditionally returning WORKER or ETH, here we always return ETH
--- a/tests/models/albert/test_albert_masked_lm.py
+++ b/tests/models/albert/test_albert_masked_lm.py
@@ -3,7 +3,7 @@
 from transformers import AutoTokenizer, AlbertForMaskedLM
 import torch
 import pytest
-from tests.utils import ModelTester
+from tests.utils import ModelTester, validate_batch_size, process_batched_logits
 
 
 class ThisTester(ModelTester):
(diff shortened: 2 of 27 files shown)


tests/models/albert/test_albert_masked_lm.py
```diff
@@ -40,15 +40,25 @@ def append_fake_loss_function(self, outputs):
...
     tester = ThisTester(model_name, mode)
-    results = tester.test_model()
-
+    results = tester.test_model(batch_size=batch_size)
+    
     if mode == "eval":
         # retrieve index of [MASK]
+        
+        results.logits = process_batched_logits(results.logits, batch_size)
+        #print(results.logits.shape)
```
jbedichekTT, line 61: remove

tests/conftest.py
```diff
@@ -70,6 +71,11 @@ def device():
     ttnn.close_device(device)
 
 
+@pytest.fixture(scope="session")
+def get_batch_size(request):
```
ayerofieiev-tt, line 75: Returned value is a `value`, not a function, so the name starting with "get" is a bit weird here. 
Wdyt?

ayerofieiev-tt: Maybe just "batch_size"

tests/models/bert/test_bert.py
```diff
@@ -35,13 +35,19 @@ def _load_inputs(self):
     ["eval"],
 )
 @pytest.mark.converted_end_to_end
-def test_bert(record_property, mode):
+def test_bert(record_property, mode, get_batch_size):
```
ayerofieiev-tt, line 38: Instead this being an option of the run here, I think it might be better to simply pass it to the ModelTester in conftest like this

```py
outputs_after = model_tester.test_model(as_ttnn=True, option=option, batch_size=batch_size)
```

tests/models/bert/test_bert.py
```diff
@@ -35,13 +35,19 @@ def _load_inputs(self):
...
 
+    batch_size = get_batch_size
+    if batch_size is not None:
+        batch_size = int(batch_size)
+    validate_batch_size(batch_size)
+
     tester = ThisTester(model_name, mode)
-    results = tester.test_model()
+    results = tester.test_model(batch_size=batch_size)
+    batch_object_inputs(tester, batch_size)  # This is necessary to avoid shape mismatch errors in tester processing
```
ayerofieiev-tt, line 50: Instead of this being here, maybe its better to have it once defined in the base `ModelTester` class somewhere around `def _load_inputs(self):`

tests/conftest.py
```diff
@@ -196,8 +202,8 @@ def compile_and_run(device, reset_torch_dynamo, request):
                     model_name, option._aten_fx_graphs, option._out_fx_graphs, option._all_inputs
                 )
 
-            if len(option._out_fx_graphs) > 0:
-                option._out_fx_graphs[0].print_tabular()
+            # if len(option._out_fx_graphs) > 0:
```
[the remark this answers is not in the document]
jbedichekTT: > restore?

Will restore

tests/models/bert/test_bert.py
jbedichekTT: > Instead this being an option of the run here, I think it might be better to simply pass it to the ModelTester in conftest like this
> 
> ```python
> outputs_after = model_tester.test_model(as_ttnn=True, option=option, batch_size=batch_size)
> ```

Will do

tests/models/bert/test_bert.py
jbedichekTT: > Instead of this being here, maybe its better to have it once defined in the base `ModelTester` class somewhere around `def _load_inputs(self):`

Will do

tests/utils.py
```diff
@@ -1,18 +1,23 @@
...
+    def __init__(self, model_name, mode, batch_size=None):
         if mode not in ["train", "eval"]:
             raise ValueError(f"Current mode is not supported: {mode}")
         self.model_name = model_name
         self.mode = mode
         self.model = self._load_model()
         self.inputs = self._load_inputs()
+        self.batch_size = batch_size
+        self.validate_batch_size()
+        self.batch_inputs()
```
ayerofieiev-tt, line 20: Should we instead make `load_inputs()` be responsible for that?

Maybe `_load_inputs()` must accept the `batch_size` parameter or respect the batch_size property of the class.
Just a thought. Open to discuss.

ayerofieiev-tt: One of the things is that we are making a shortcut right now, just cloning input. But ideally, we should have N different inputs batched into a single run.
We can have a shortcut now, but from test arch point of view, it is great if its easy to later make an improvement for a given model.

jbedichekTT: > Should we instead make `load_inputs()` be responsible for that?
> 
> Maybe `_load_inputs()` must accept the `batch_size` parameter or respect the batch_size property of the class. Just a thought. Open to discuss.

I think it might read a bit cleaner, but I think this way is better because there is less code, so its easier to change in the future.

PR #806, opened 2025-03-01
Integrate custom TTNN extension for Pytorch

kevinwuTT:
### Problem description
We want to be able to cache the model weights by calling`model.to("ttnn_device")` which should transfer the data to device before running the model.

We can utilize the [open_registration_extension](https://github.com/pytorch/pytorch/blob/v2.2.1/test/cpp_extensions/open_registration_extension.cpp) (or a more standalone example: [pytorch_open_registration_example](https://github.com/bdhirsh/pytorch_open_registration_example)) to intercept calls to `torch_tensor.to(device)`. 

### What's changed
We can subclass`TensorImpl` in order to hold our own custom data. In other words, we can copy data to device in advanced and store a reference to a TTNN tensor inside a Torch.tensor. That way we can avoid calling `ttnn.from_torch` when we encounter a Torch.tensor in our compiled graphs. This should be useful when calling the model for multiple iterations. 

Build changes:
The bulk of the implementation has to be done on C++, thus we will have to compile and link against tt-metal and Pytorch libraries. Currently, tt-metal uses `clang++-17` with `--std=c++20` and `--std=libc++`. However, the standard Pytorch wheels use `g++` and `--std=c++17` and `--std=stdlibc++`. The standard C++ libraries are incompatible, thus we have to recompile Pytorch. 

TODO:

- [ ] Integrate into conftest.py and add a commandline option. Also make sure to not load the extension at all (from import) if feature is disabled
- [ ] Make `torch.utils.rename_privateuse1_backend("ttnn")` more elegant
- [ ] Set up custom build of Pytorch with clang
- [ ] Can we link with ttnn wheel?
- [ ] Clean up `TtnnDeviceMode` class
- [ ] Clean up unused headers
- [ ] Review if we really need a custom `at::Allocator` class
- [ ] Clean up Logging in C++ files
- [ ] Clean up unused "kernels" in open_registration_extension.cpp
- [ ] Review accessing data via `unsafeGetTensorImpl`

### DIFF
--- a/tests/cpp_extension/test_cpp_extension.py
+++ b/tests/cpp_extension/test_cpp_extension.py
@@ -1,0 +1,165 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torchvision import transforms, datasets
+from torch.utils.data import DataLoader
+import ttnn
+import torch_ttnn
+from torch_ttnn.cpp_extension.custom_device_mode import ttnn_module, enable_ttnn_device
+import pytest
+
+from transformers import AutoTokenizer, AutoModelForQuestionAnswering
+
+import logging
+import sys
+
+@pytest.mark.parametrize(
+    "input_shape",
+    ((32, 1, 3, 3), (32,)),
+)
+def test_cpp_extension(device, input_shape):
+    torch.utils.rename_privateuse1_backend('ttnn')
+
+    # in pytest the device has already been initialized before this call
+    # so instead we can wrap this around the custom device
+    ttnn_device = ttnn_module.custom_device_from_ttnn(device)
+
+    logging.info("Creating bfloat tensor from -1 to 1")
+    torch_tensor = torch.empty(input_shape, dtype = torch.bfloat16).uniform_(-1, 1)
+    print(torch_tensor)
+    torch_tensor_abs = torch.abs(torch_tensor)
+    print(torch_tensor_abs)
+
+    logging.info("Transferring to ttnn")
+    torch_ttnn_tensor = torch_tensor.to(ttnn_device)
+
+    logging.info("get underlying ttnn tensor")
+    ttnn_tensor = ttnn_module.get_ttnn_tensor(torch_ttnn_tensor)
+
+    logging.info("Running abs on ttnn")
+    ttnn_tensor = ttnn.abs(ttnn_tensor)
+
+    logging.info("calling to_torch")
+    ttnn_to_torch = ttnn.to_torch(ttnn_tensor)
+    print(ttnn_to_torch)
+    
+    
+    assert torch.allclose(torch_tensor_abs, ttnn_to_torch, rtol=0.1, atol=0.1)
+
+    # logging.info("Closing device")
+    # ttnn_module.close_custom_device(ttnn_device)
+
+def test_bert_with_cpp_extension(device):
+    model_name = "phiyodr/bert-large-finetuned-squad2"
+    tokenizer = AutoTokenizer.from_pretrained(model_name, padding_side="left", torch_dtype=torch.bfloat16)
+    m = AutoModelForQuestionAnswering.from_pretrained(model_name, torch_dtype=torch.bfloat16)
+    context = 'Johann Joachim Winckelmann was a German art historian and archaeologist. He was a pioneering Hellenist who first articulated the difference between Greek, Greco-Roman and Roman art. "The prophet and founding hero of modern archaeology", Winckelmann was one of the founders of scientific archaeology and first applied the categories of style on a large, systematic basis to the history of art. '
+    question = "What discipline did Winkelmann create?"
+    inputs = tokenizer.encode_plus(
+        question,
+        context,
+        add_special_tokens=True,
+        return_tensors="pt",
+        max_length=256,
+        padding="max_length",
+        truncation=True,
+    )
+
+    option = torch_ttnn.TorchTtnnOption(
+                    device=device,
+                    gen_graphviz=False,
+                    run_mem_analysis=False,
+                    metrics_path=model_name,
+                    verbose=True,
+                )
+
+    # custom device
+    torch.utils.rename_privateuse1_backend('ttnn')
+    ttnn_device = ttnn_module.custom_device_from_ttnn(device)
+    
+    # clone input_ids on cpu since this the data transfer is somehow inplace?
+    input_ids = inputs.input_ids.clone()
+    
+    inputs = inputs.to(ttnn_device)
+    # modules are inplace, tensors are not
+    m.to(ttnn_device)
+
+    model = torch.compile(m, backend=torch_ttnn.backend, options=option)
+    outputs = model(**inputs)
+    
+    # Helper function to decode output to human-readable text
+    def decode_output(outputs):
+        response_start = torch.argmax(outputs.start_logits)
+        response_end = torch.argmax(outputs.end_logits) + 1
+        response_tokens = input_ids[0, response_start:response_end]
+        return tokenizer.decode(response_tokens)
+
+    print("finished:")
+    print(outputs)
+    answer = decode_output(outputs)
+
+    print(
+        f"""
+    model_name: {model_name}
+    input:
+        context: {context}
+        question: {question}
+    answer: {answer}
+    """
+    )
+
+# adapted from https://github.com/pytorch/examples/blob/main/mnist/main.py
+class MnistModel(torch.nn.Module):
+    def __init__(self):      
+        super(MnistModel, self).__init__()
+        self.conv1 = nn.Conv2d(1, 32, 3, 1)
+        self.conv2 = nn.Conv2d(32, 64, 3, 1)
+        self.dropout1 = nn.Dropout(0.25)
+        self.dropout2 = nn.Dropout(0.5)
+        self.fc1 = nn.Linear(9216, 128)
+        self.fc2 = nn.Linear(128, 10)
+
+    def forward(self, x):
+        x = self.conv1(x)
+        x = F.relu(x)
+        x = self.conv2(x)
+        x = F.relu(x)
+        x = F.max_pool2d(x, 2)
+        x = self.dropout1(x)
+        x = torch.flatten(x, 1)
+        x = self.fc1(x)
+        x = F.relu(x)
+        x = self.dropout2(x)
+        x = self.fc2(x)
+        x = F.log_softmax(x, dim=1)
+        return x
+    
+def test_mnist_with_cpp_extension(device):   
+    model_name = "Mnist"
+    transform = transforms.Compose([transforms.ToTensor()])
+    test_dataset = datasets.MNIST(root="./data", train=False, transform=transform, download=True)
+    dataloader = DataLoader(test_dataset, batch_size=1)
+    test_input, _ = next(iter(dataloader))
+    test_input = test_input.to(torch.bfloat16)
+
+    # Copy weights and biases to ttnn
+    torch.utils.rename_privateuse1_backend('ttnn')
+    ttnn_device = ttnn_module.custom_device_from_ttnn(device)
+    
+
+    
+    option = torch_ttnn.TorchTtnnOption(
+                    device=device,
+                    gen_graphviz=False,
+                    run_mem_analysis=False,
+                    metrics_path=model_name,
+                    verbose=True,
+                )
+
+    model = MnistModel()
+    model = model.to(torch.bfloat16)
+    test_input = test_input.to(ttnn_device)
+    model.to(ttnn_device)
+            
+    model = torch.compile(model, backend=torch_ttnn.backend, options=option)
+    results = model(test_input)
--- a/tests/cpp_extension/test_cpp_extension.py
+++ b/tests/cpp_extension/test_cpp_extension.py
@@ -7,47 +7,64 @@ import ttnn
 import torch_ttnn
 from torch_ttnn.cpp_extension.custom_device_mode import ttnn_module, enable_ttnn_device
 import pytest
+import time
 
 from transformers import AutoTokenizer, AutoModelForQuestionAnswering
 
 import logging
 import sys
 
+
 @pytest.mark.parametrize(
     "input_shape",
-    ((32, 1, 3, 3), (32,)),
+    ((32, 1, 3, 3), (1, 32)),
 )
-def test_cpp_extension(device, input_shape):
-    torch.utils.rename_privateuse1_backend('ttnn')
+@pytest.mark.parametrize(
+    "dtype",
+    (torch.bfloat16, torch.int32),
+)
+def test_cpp_extension(device, input_shape, dtype):
+    torch.utils.rename_privateuse1_backend("ttnn")
 
     # in pytest the device has already been initialized before this call
     # so instead we can wrap this around the custom device
     ttnn_device = ttnn_module.custom_device_from_ttnn(device)
 
     logging.info("Creating bfloat tensor from -1 to 1")
-    torch_tensor = torch.empty(input_shape, dtype = torch.bfloat16).uniform_(-1, 1)
+    if dtype == torch.bfloat16:
+        torch_tensor = torch.empty(input_shape, dtype=dtype).uniform_(-1, 1)
+    elif dtype == torch.int32:
+        torch_tensor = torch.randint(-1000, 1000, input_shape)
+        torch_tensor = torch_tensor.to(torch.int32)
+    else:
+        raise Exception(f"{dtype} not being tested at this time")
     print(torch_tensor)
-    torch_tensor_abs = torch.abs(torch_tensor)
-    print(torch_tensor_abs)
 
     logging.info("Transferring to ttnn")
     torch_ttnn_tensor = torch_tensor.to(ttnn_device)
 
-    logging.info("get underlying ttnn tensor")
+    logging.info("Get underlying ttnn tensor")
     ttnn_tensor = ttnn_module.get_ttnn_tensor(torch_ttnn_tensor)
 
-    logging.info("Running abs on ttnn")
-    ttnn_tensor = ttnn.abs(ttnn_tensor)
+    # Compare output of abs op for bfloat16 dtype since ttnn.abs does not support int
+    if dtype == torch.bfloat16:
+        torch_out = torch.abs(torch_tensor)
+        print(torch_out)
+
+        logging.info("Running abs on ttnn")
+        ttnn_tensor = ttnn.abs(ttnn_tensor)
+    elif dtype == torch.int32:
+        torch_out = torch_tensor
+    else:
+        raise Exception(f"{dtype} not being tested at this time")
 
     logging.info("calling to_torch")
     ttnn_to_torch = ttnn.to_torch(ttnn_tensor)
-    print(ttnn_to_torch)
-    
-    
-    assert torch.allclose(torch_tensor_abs, ttnn_to_torch, rtol=0.1, atol=0.1)
 
-    # logging.info("Closing device")
-    # ttnn_module.close_custom_device(ttnn_device)
+    print(ttnn_to_torch)
+
+    assert torch.allclose(torch_out, ttnn_to_torch, rtol=0.1, atol=0.1)
+
 
 def test_bert_with_cpp_extension(device):
     model_name = "phiyodr/bert-large-finetuned-squad2"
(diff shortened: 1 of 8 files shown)


torch_ttnn/cpp_extension/custom_device_mode.py
```diff
@@ -0,0 +1,102 @@
+import torch
+import torch.utils.cpp_extension
+from torch.overrides import TorchFunctionMode
+import os
+from pathlib import Path
+import glob
+import logging
+
+assert os.environ.get("TT_METAL_HOME") is not None
+tt_metal_home = Path(os.environ["TT_METAL_HOME"])
```
ayerofieiev-tt, line 10: we probably can fetch from ttnn.__file__ location

torch_ttnn/cpp_extension/TtnnTensorImpl.hpp
```diff
@@ -0,0 +1,128 @@
+#pragma once
+
+#include "ttnn/tensor/tensor.hpp"
+#include "extension_utils.hpp"
+#include <iostream>
+#include <string.h>
+
+namespace at {
+
+struct TtnnTensorImpl : public TensorImpl {
```
ayerofieiev-tt, line 10: lets split into .hpp and .cpp

torch_ttnn/cpp_extension/TtnnTensorImpl.hpp
```diff
@@ -0,0 +1,128 @@
...
+    LOGGING("");
+    return ttnn_tensor_;
+  }
+
+  void set_ttnn_tensor(const ttnn::Tensor& tensor) {
+    ttnn_tensor_ = tensor;
+  }
+
+  /**
+   * Return a TensorImpl that is a shallow-copy of this TensorImpl.
```
ayerofieiev-tt, line 58: What `shallow` means in this case?

tests/cpp_extension/test_cpp_extension.py
ayerofieiev-tt, line 1: lets split into multiple files

tests/cpp_extension/test_cpp_extension.py
```diff
@@ -0,0 +1,193 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torchvision import transforms, datasets
+from torch.utils.data import DataLoader
+import ttnn
+import torch_ttnn
+from torch_ttnn.cpp_extension.custom_device_mode import ttnn_module, enable_ttnn_device
```
ayerofieiev-tt, line 8: no need to call `enable_ttnn_device()`?

tests/cpp_extension/test_cpp_extension.py
```diff
@@ -0,0 +1,193 @@
...
+@pytest.mark.parametrize(
+    "input_shape",
+    ((32, 1, 3, 3), (1, 32)),
+)
+@pytest.mark.parametrize(
+    "dtype",
+    (torch.bfloat16, torch.int32),
+)
+def test_cpp_extension(device, input_shape, dtype):
+    torch.utils.rename_privateuse1_backend("ttnn")
```
ayerofieiev-tt, line 27: can this be hidden a bit?
user can call some init, that init can do these things if needed. 
this line is official but pretty cryptic and would be good to avoid

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,628 @@
...
+  auto dtype = c10::scalarTypeToTypeMeta(dtype_opt.value());
+  LOGGING("dtype: ", dtype);
+  at::Device device = device_opt.value();
+  LOGGING("device: ", device);
+  TtnnGuard device_guard(device);
+  IDevice* ttnn_device = device_guard.get_ttnn_device();
+  LOGGING("size: ", size);
+  ttnn::SmallVector<uint32_t> small_vector(size.begin(), size.end());
+
+  // TODO: This preallocated data doesn't get used, but the metadata does. Needs fixing.
```
ayerofieiev-tt, line 257: means there is a reduntant copy?

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,628 @@
...
+    /*resizeable=*/true);
+
+  // Store ttnn tensor to torch tensor impl
+  auto tensor_ret = at::detail::make_tensor<at::TtnnTensorImpl>(private_use_ks, dtype, device, tensor, storage_impl);
+  LOGGING("");
+
+  return tensor_ret;
+}
+
+at::Tensor & custom_fill__scalar(at::Tensor & self, const at::Scalar & value) {
```
ayerofieiev-tt, line 281: need to remove else its error prone?

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,628 @@
...
+
+      // Debug only: Convert torch tensor to vector
+      auto self_float = self.to(at::kFloat);
+      std::vector<float> v(self_float.data_ptr<float>(), self_float.data_ptr<float>() + self_float.numel());
+      LOGGING("src_ten: ", v);
+
+      // Debug only: Convert ttnn tensor on cpu to vector
+      auto src_cpu_vector = src_cpu.to_vector<float>();
+      LOGGING("src_cpu: ", src_cpu_vector);
+      vector_compare(v, src_cpu_vector);
```
ayerofieiev-tt, line 352: if debug only, please hide under ifdef

torch_ttnn/cpp_extension/TtnnCustomAllocator.h
```diff
@@ -0,0 +1,23 @@
+#include <c10/core/Allocator.h>
+
+#include "extension_utils.hpp"
+
+// A dummy allocator for the custom device, that secretly uses the CPU
+// TODO: Do we have to implement this for the use of TTNN?
+struct TtnnCustomAllocator final : at::Allocator {
+    TtnnCustomAllocator() = default;
+    at::DataPtr allocate(size_t nbytes) const override {
+        LOGGING("");
```
ayerofieiev-tt, line 10: lets remove?

torch_ttnn/cpp_extension/TtnnGuard.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+// Represents the current "active" device.
+// The device guard registered below is meant to show how a backend
+// can integrate custom device guard with pytorch.
+// For something like cuda this represents the current active cuda device,
+// which is directly set using the cuda API calls cudaGetDevice/cudaSetDevice.
+static uint16_t CURR_DEVICE = -1;
+
+// Create and register a dummy device guard.
+struct TtnnDeviceGuard final : public c10::impl::DeviceGuardImplInterface {
```
ayerofieiev-tt, line 31: I see things split into .hpp/.cpp but not fully

torch_ttnn/cpp_extension/TtnnTensorImpl.hpp
```diff
@@ -1,128 +1,62 @@
...
+
+    void set_sizes_and_strides(const IntArrayRef& int_array_ref);
+
+    void set_sizes_and_strides_as(const at::Tensor& the_template);
+
+    ttnn::Tensor get_ttnn_tensor();
+
+    void set_ttnn_tensor(const ttnn::Tensor& tensor);
+
+    ttnn::Shape get_logical_shape();
```
ayerofieiev-tt, line 28:
```suggestion
    ttnn::Shape get_logical_shape() const;
```

torch_ttnn/cpp_extension/TtnnTensorImpl.hpp
```diff
@@ -1,128 +1,62 @@
...
+        const caffe2::TypeMeta data_type,
+        c10::Device device,
+        const at::IntArrayRef& size,
+        const c10::Storage& storage);
+
+    void set_sizes_and_strides(const IntArrayRef& int_array_ref);
+
+    void set_sizes_and_strides_as(const at::Tensor& the_template);
+
+    ttnn::Tensor get_ttnn_tensor();
```
ayerofieiev-tt, line 24:
```suggestion
    ttnn::Tensor get_ttnn_tensor() const;
```

torch_ttnn/cpp_extension/TtnnTensorImpl.hpp
```diff
@@ -1,128 +1,62 @@
...
+    TtnnTensorImpl(
+        at::DispatchKeySet key_set,
+        const caffe2::TypeMeta data_type,
+        c10::Device device,
+        const at::IntArrayRef& size,
+        const c10::Storage& storage);
+
+    void set_sizes_and_strides(const IntArrayRef& int_array_ref);
+
+    void set_sizes_and_strides_as(const at::Tensor& the_template);
```
ayerofieiev-tt, line 22: add short doc

torch_ttnn/cpp_extension/TtnnTensorImpl.hpp
```diff
@@ -1,128 +1,62 @@
...
+     * `allow_tensor_metadata_change_`, see NOTE [ TensorImpl Shallow-Copying ].
+     */
+    void shallow_copy_from(const c10::intrusive_ptr<TensorImpl>& impl) override;
+
+private:
     ttnn::Tensor ttnn_tensor_;
-    // TODO: Debug only, should probably remove as it might be costly
-    std::string ttnn_tensor_string_;
+    ttnn::Shape logical_shape_;
+    bool is_empty_initialized_;
```
ayerofieiev-tt, line 59:
```suggestion
    bool is_empty_initialized_ = false;
```

tests/cpp_extension/test_cpp_extension.py
kevinwuTT: Done. Moved to module initialization so it should only be done once.

torch_ttnn/cpp_extension/open_registration_extension.cpp
kevinwuTT: Latest refactoring will use a runtime env variable `DEBUG_CPP_EXT` to enable/disable debug prints and checks.

torch_ttnn/cpp_extension/TtnnTensorImpl.hpp
kevinwuTT: This function is based on https://github.com/pytorch/pytorch/blob/16560d4e8f8a33844d3bc629ecc6ff2c4c290a54/c10/core/TensorImpl.h#L1991. From my understanding, this is used when two variables point to the same tensor metadata but with different autograd history. So this function allows handling of the metadata without copying the tensor data itself. In addition, I also copied the TTNN Tensor variable, but my understanding i that this doesn't make a deep copy, just increases the reference count.

torch_ttnn/cpp_extension/extension_utils.hpp
```diff
@@ -0,0 +1,48 @@
...
+    std::transform(src_vector.begin(), src_vector.end(), std::back_inserter(dst_vector), [](const uint32_t value) {
+        return static_cast<int>(value);
+    });
+}
+
+#define GET_OPTIONAL_OR_ASSERT(optional)             \
+    ({                                               \
+        TORCH_INTERNAL_ASSERT(optional.has_value()); \
+        optional.value();                            \
+    })
```
ayerofieiev-tt, line 48: I recommend to replace

```cpp
template<typename T>
T getOptionalOrAssert(const std::optional<T>& opt) {
    TORCH_INTERNAL_ASSERT(opt.has_value());
    return opt.value();
}
```

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,362 @@
...
+            auto ttnn_vector = std::get<std::vector<float>>(tensor_to_vector(ttnn_tensor));
+
+            vector_compare(torch_vector, ttnn_vector);
+        } else if (ttnn_dtype == ttnn::DataType::UINT32) {
+            auto torch_vector = std::get<std::vector<int>>(tensor_to_vector(torch_tensor));
+            auto ttnn_vector = std::get<std::vector<int>>(tensor_to_vector(ttnn_tensor));
+
+            vector_compare(torch_vector, ttnn_vector);
+        }
+    }
```
ayerofieiev-tt, line 96:
```suggestion
    if (!DEBUG_CPP_EXT) {
        return
    }

    auto ttnn_dtype = ttnn_tensor.dtype();    
    if (ttnn_dtype == ttnn::DataType::BFLOAT16) {
        auto torch_vector = std::get<std::vector<float>>(tensor_to_vector(torch_tensor));
        auto ttnn_vector = std::get<std::vector<float>>(tensor_to_vector(ttnn_tensor));

        vector_compare(torch_vector, ttnn_vector);
    } else if (ttnn_dtype == ttnn::DataType::UINT32) {
        auto torch_vector = std::get<std::vector<int>>(tensor_to_vector(torch_tensor));
        auto ttnn_vector = std::get<std::vector<int>>(tensor_to_vector(ttnn_tensor));

        vector_compare(torch_vector, ttnn_vector);
    }
```

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,362 @@
...
+        std::vector<int> src_vector_int;
+        convert_vector_from_uint32_to_int(src_vector_int, ttnn_vector);
+        LOGGING("src_dev: ", src_vector_int);
+        return src_vector_int;
+    } else {
+        TORCH_INTERNAL_ASSERT(false);
+    }
+}
+
+void compare_torch_and_ttnn_tensors(const at::Tensor& torch_tensor, const ttnn::Tensor& ttnn_tensor) {
```
ayerofieiev-tt, line 81:
```suggestion
void assert_equal(const at::Tensor& torch_tensor, const ttnn::Tensor& ttnn_tensor) {
```

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,362 @@
...
+            auto ttnn_vector = std::get<std::vector<int>>(tensor_to_vector(ttnn_tensor));
+
+            vector_compare(torch_vector, ttnn_vector);
+        }
+    }
+}
+
+namespace {
+
+void abs_kernel(at::TensorIteratorBase& iter) {
```
ayerofieiev-tt, line 101: Should we make A+B work natively in this PR?
Else, I am not sure whats the value of having an empty non working function

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+}  // namespace
+
+namespace at::native {
+REGISTER_PRIVATEUSE1_DISPATCH(abs_stub, &abs_kernel);
+}
+
+// Register custom allocator. Only used to create dummy Torch tensor object.
+static TtnnCustomAllocator ttnn_custom_alloc;
+REGISTER_ALLOCATOR(c10::DeviceType::PrivateUse1, &ttnn_custom_alloc);
```
ayerofieiev-tt, line 113: consider this instead 
```suggestion
TtnnCustomAllocator get_custom_ttnn_allocator(){
    static TtnnCustomAllocator ttnn_custom_alloc;
 }
REGISTER_ALLOCATOR(c10::DeviceType::PrivateUse1, &get_custom_ttnn_allocator());
```

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,362 @@
...
+    auto out_ttnn_tensor = out_tensor_impl->get_ttnn_tensor();
+    out_tensor_impl->set_ttnn_tensor(result);
+
+    return out;
+}
+
+// This macro registers the kernels to the PyTorch Dispatcher.
+// More details on the dispatcher can be found at
+// http://blog.ezyang.com/2020/09/lets-talk-about-the-pytorch-dispatcher/.
+TORCH_LIBRARY_IMPL(aten, PrivateUse1, m) {
```
ayerofieiev-tt, line 313: can we give an alias to `PrivateUse1`? Like `TTNN_Backend`?

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+// ==============================================
+// ============== PYBIND FUNCTIONS ==============
+// ==============================================
+
+// This function can be used when the TTNN device is initialized separately,
+// for example, `device = ttnn.open_device(device_index = 0)`. Pass that
+// device object to this function so that the cpp extension can use it.
+c10::Device as_torch_device(IDevice* ttnn_device) {
+    LOGGING("");
```
ayerofieiev-tt, line 328: do we need this? what do you see when this is printed?

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,362 @@
...
+    // TODO: Fix the index
+    auto device = c10::Device(c10::DeviceType::PrivateUse1, 0);
```
ayerofieiev-tt, line 330: what 0 means here?

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+// Manually closes the torch and ttnn device
+void close_torch_device(c10::Device device) {
+    LOGGING("");
+    TtnnGuard device_guard(device);
+    IDevice* ttnn_device = device_guard.get_ttnn_device();
+    // TODO: Perform better error handling
+    TORCH_INTERNAL_ASSERT(ttnn_device != nullptr);
+    ttnn::close_device(*ttnn_device);
+    ttnn_device = nullptr;
```
ayerofieiev-tt, line 345: need to add this?
```cpp
TtnnGuard::ttnn_device = nullptr;
```

torch_ttnn/cpp_extension/open_registration_extension.cpp
```diff
@@ -0,0 +1,362 @@
...
+}
+
+// Get the underlying TTNN tensor from a Torch tensor
+ttnn::Tensor get_ttnn_tensor(at::Tensor& tensor) {
+    LOGGING("");
+    // TODO: Check if this cast fails
```
ayerofieiev-tt, line 351: Might actually fail if you give it torch cpu tensor?

torch_ttnn/cpp_extension/ttnn_device_mode.py
```diff
@@ -0,0 +1,104 @@
...
+working_file_path = Path(os.path.realpath(__file__))
+working_directory = working_file_path.parents[0]
+
+ttnn_module = torch.utils.cpp_extension.load(
+    name="ttnn_device_extension",
+    sources=[
+        str(working_directory / "open_registration_extension.cpp"),
+        str(working_directory / "TtnnCustomAllocator.cpp"),
+        str(working_directory / "TtnnGuard.cpp"),
+        str(working_directory / "TtnnTensorImpl.cpp"),
```
ayerofieiev-tt, line 54: maybe programmatically list all `.cpp` files in the dir and use that list?

tests/cpp_extension/test_cpp_extension_functionality.py
```diff
@@ -0,0 +1,77 @@
...
+    ((32, 1, 3, 3), (1, 32), (4,)),
+)
+@pytest.mark.parametrize(
+    "dtype",
+    (torch.bfloat16, torch.int, torch.long),
+)
+def test_cpp_extension(device, input_shape, dtype):
+    # in pytest the device has already been initialized before this call
+    # so instead we can wrap this around the custom device
+    ttnn_device = ttnn_module.as_torch_device(device)
```
ayerofieiev-tt, line 20:
```suggestion
def test_cpp_extension(ttnn_device, input_shape, dtype):
    # in pytest the device has already been initialized before this call
    # so instead we can wrap this around the custom device
    device = ttnn_module.as_torch_device(ttnn_device)
```

tests/cpp_extension/test_cpp_extension_functionality.py
```diff
@@ -0,0 +1,77 @@
...
+        logging.info("Running abs on ttnn")
+        ttnn_tensor = ttnn.abs(ttnn_tensor)
+    elif dtype == torch.int or dtype == torch.long:
+        torch_out = torch_tensor
+    else:
+        raise Exception(f"{dtype} not being tested at this time")
+
+    logging.info("calling to_torch")
+    ttnn_to_torch = ttnn.to_torch(ttnn_tensor)
+    # cast back to dtype because ttnn always return uint32_t for integers
```
ayerofieiev-tt, line 52: it would be a good test to show that long values are truncated here

tests/cpp_extension/test_cpp_extension_functionality.py
kevinwuTT: This is a dilemma because the `device` parameter is the fixture name from `conftest.py` so we can't really rename it.

torch_ttnn/cpp_extension/open_registration_extension.cpp
kevinwuTT: This is an index used for multi-devices for Torch device API. I'll see if I can generalize it for the time being in preparation for supporting multi-devices.

torch_ttnn/cpp_extension/open_registration_extension.cpp
kevinwuTT: The LOGGING macro currently prints the file, function,line number, and the string. It's only enabled if an environment variable `DEBUG_CPP_EXT` is defined during runtime. Otherwise, it's a nop.

torch_ttnn/cpp_extension/open_registration_extension.cpp
kevinwuTT: The key `PrivateUse1` is hardcoded into Pytorch. But the idea is that any out-of-source backends like ours can use this key to make our implementation. Then we call `torch.rename_privateuse1_backend("ttnn")` to rename it to our own so that in the user-facing side, they can just call `tensor.to("ttnn")`. [Rename PrivateUse1 to a custom name for the new backend](https://pytorch.org/tutorials/advanced/privateuseone.html#rename-privateuse1-to-a-custom-name-for-the-new-backend) In other words, we don't really rename this key in our implementation, but we can rename it for the user. I do have the rename call under the module that initializes this extension: `ttnn_device_mode.py`.

torch_ttnn/cpp_extension/open_registration_extension.cpp
kevinwuTT: This is a strange thing that I've encountered when dealing with some ops like `abs`. When intercepting this op, I already have another function defined in this file called `at::Tensor& custom_abs_out(const at::Tensor& self, at::Tensor& out)` which gets called first. I can already call `ttnn::abs` at this point, so I technically don't need the `void abs_kernel(at::TensorIteratorBase& iter)` function. However, if I don't define this second function, I get a runtime error saying my backend is missing an implementation. So that's the reason I have that function defined even if it doesn't do anything. I can try to dig a bit deeper to see if I'm missing a way to bypass that.

PR #812, opened 2025-03-04
`aten.native_layer_norm.default` lowering modification to include custom compute kernel config

amalbasaTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-metal/issues/18345)

### Problem description
ttnn.layer_norm found in albert_v2 models have low accuracy in some occurrences. Adding compute kernel config with `fp32_dest_acc_en` enabled fixes that issue. 

### What's changed
1) Added `TtnnComputeKernelConfig` helper class.
2) Added `compute_kernel_config` registration to torch.fx graph module.
3) In to_tt_pass.py modified lowering of `torch.ops.`aten.native_layer_norm.default` to include custom compute kernel config.
4) In tools/generate_op_accuracy_tests.py, expaned `_build_code_from_aten_ttnn_graphs` to include kernel config at the beginning of forward `function.`


kevinwuTT:
I think one cleaner approach when dealing with more complicated calls to `ttnn` functions is to create a wrapper like this: https://github.com/tenstorrent/pytorch2.0_ttnn/blob/main/torch_ttnn/passes/lowering/target_wrappers.py. That way, we don't have to modify `tools/generate_op_accuracy_tests.py` because it should automatically grab the wrapper function.

amalbasaTT:
@kevinwuTT I assumed that in `tools/generate_op_accuracy_tests.py` we should add variables which are common for most of ops in `forward` function of generated code. `compute_kernel_config` corresponds to that, and I'm not sure how will wrapping layer_norm make it cleaner (if i add `compute_kernel_config` to layer_norm wrapper, there will be as many same `compute_kernel_configs` as there is layer_norms inside graph).

PR #822, opened 2025-03-12
Refactor AddDataMovePass

jmalone-tt:
I tried to separate changes into atomic commits. That is, each commit contains a (mostly) related set of changes, and I would expect tests to pass after any given commit.

There were a handful of functions that I was able to cut down significantly, but I would like to remove more of the special cases in the future.

I have been looking at most of this code long enough that I am losing a sense of what is unreadable, so hoping to get some other opinions.

I also considered making some bigger changes, but am not sure if it is worthwhile. Please let me know your thoughts in comments:
1. Changing variable name `spec` to something more descriptive like `desired_alignment`
2. Changing control flow. Right now each of the transitions (Torch->Ttnn, Ttnn->Torch, and Ttnn->Ttnn) flow through the same functions. Most of these functions case on transition type, so it could be cleaner to have a function per transition type that handles multiple steps (getting aligned spec and creating aligned nodes). I didn't do this yet because we use a dictionary to cache and reuse previous aligned nodes, which this would likely break.

### DIFF
--- a/torch_ttnn/passes/lowering/add_data_move_pass.py
+++ b/torch_ttnn/passes/lowering/add_data_move_pass.py
@@ -294,6 +294,11 @@ class NodeInputAligner:
         if input_node.target == ttnn.sharded_to_interleaved and input_node.args[0].target == ttnn.max_pool2d:
             spec.layout = TtnnTileLayout
 
+        # be overly cautious and convert to tile layout. These could already be tilized
+        # TODO: only insert layout if needed
+        if input_node.target in [ttnn.ones, ttnn.ones_like, ttnn.zeros, ttnn.zeros_like]:
+            spec.layout = TtnnTileLayout
+
         # legalize to the default layout and device
         if input_node.target in TTNN_ROW_LAYOUT_OPS:
             spec.layout = TtnnTileLayout
--- a/torch_ttnn/passes/lowering/add_data_move_pass.py
+++ b/torch_ttnn/passes/lowering/add_data_move_pass.py
@@ -355,16 +355,7 @@ class NodeInputAligner:
         if input_node.target == getitem and input_node.args[0].target == ttnn.split:
             spec.layout = TtnnTileLayout
         # legalize to the default layout and device
-        if input_node.target in TTNN_LAYOUT_CHANGE_OPS.union(
-            set(
-                [
-                    target_wrappers.repeat,
-                    target_wrappers.roll,
-                    target_wrappers.stack,
-                    ttnn.concat,
-                ]
-            )
-        ):
+        if input_node.target in TTNN_LAYOUT_CHANGE_OPS:
             spec.layout = TtnnTileLayout
             spec.device = TtnnDevice
         return spec
--- a/torch_ttnn/passes/lowering/add_data_move_pass.py
+++ b/torch_ttnn/passes/lowering/add_data_move_pass.py
@@ -159,13 +159,15 @@ TTNN_POOL_OPS = [
     ttnn.max_pool2d,
 ]
 
-TTNN_LAYOUT_CHANGE_OPS = set(
+TTNN_ROW_LAYOUT_OPS = set(
     [
-        ttnn.reshape,
         ttnn.slice,
+        target_wrappers.roll,
         ttnn.argmax,
     ]
 )
+
+TTNN_HOST_ONLY_OPS = set()
 
 
 # For operations limitations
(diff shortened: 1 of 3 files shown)


ayerofieiev-tt (commented):
Screenshot 2025-03-12 at 2 08 51 PM
good sign

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -381,47 +354,46 @@ def _get_align_spec(self, node, input_node, input_site, input_site_type: InputSi
...
+        new_input_node = self.graph.call_function(ttnn.to_torch, (spec.input_node,), {"dtype": spec.dtype})
+
+        if spec.input_node.meta is not None:
+            new_input_node.meta = spec.input_node.meta.copy()
+        if "original_input_variations" in new_input_node.meta:
+            new_input_node.meta["original_input_variations"] = None
+
+        return new_input_node
+
+    def _change_layout(self, spec):
```
ayerofieiev-tt, line 378: Any idea why `device` would be null?

ayerofieiev-tt (commented):
There are lots of `to_torch` calls here. 
Would be good to know why we need them.
I assume thats because back in a time some layout changes did not work on device.
That may be not the case anymore.
And if thats still the case, we need to fire a ticket (and maybe fix ourself if TM team can't help fast) to handle in tt_metal.

ayerofieiev-tt (commented):
In general looks like an improvement. Thank you!

torch_ttnn/passes/lowering/add_data_move_pass.py
jmalone-tt: Based on my understanding, the `None` values basically mean that we don't explicitly want to set the device/layout/dtype to a given value. Concretely, Ttnn->Ttnn layout changes start with `device=None` and will have the device updated to TtnnDevice if the input node is in the set of TTNN_LAYOUT_CHANGE_OPS or the node is one of the special cases handled in `_align_special_cases`.
@kevinwuTT might be able to clarify if my understanding is correct though

kevinwuTT: Yeah, in the past there were some ops like ttnn.reshape that did not work when the tensor is on device for specific shapes. From my understanding, `spec.device == None` means this tensor should not be on device for this op to work.

kevinwuTT:
I wonder if removing the need for `TTNN_LAYOUT_CHANGE_OPS` as well as individual conditions for ops in that list can be part of this refactoring. For example, if I remove the following, BERT test will run to completion and the amount of `to_layout/to_device` ops have been significantly reduced. (Although I am seeing some accuracy issues.) But aside from that, this is somewhat promising since that means a lot of the reshape issues may have been fixed. We can try with the other ops in this list too.

```diff
diff --git a/torch_ttnn/passes/lowering/add_data_move_pass.py b/torch_ttnn/passes/lowering/add_data_move_pass.py
index 42e6c927..598592ab 100644
--- a/torch_ttnn/passes/lowering/add_data_move_pass.py
+++ b/torch_ttnn/passes/lowering/add_data_move_pass.py
@@ -161,7 +161,6 @@ TTNN_POOL_OPS = [
 
 TTNN_LAYOUT_CHANGE_OPS = set(
     [
-        ttnn.reshape,
         ttnn.slice,
         ttnn.argmax,
     ]
@@ -329,14 +328,6 @@ class NodeInputAligner:
             # TODO(#417, tt-metal#15893): weight currently needs to be on host and can't be moved to device first
             spec.layout = TtnnRowMajorLayout
             spec.device = "host"
-        if (
-            node.target == ttnn.reshape
-            and hasattr(spec.input_node, "meta")
-            and "val" in spec.input_node.meta
-            and hasattr(spec.input_node.meta["val"], "dtype")
-            and spec.input_node.meta["val"].dtype in [torch.int32, torch.int64]
-        ):
-            spec.dtype = TtnnUint32
         if node.target == ttnn.permute and len(node.meta["val"].size()) > 4:
             # TODO(tt-metal#16188): to_layout on device fails for > 4D
             # Otherwise ttnn.permute can do to_layout internally and this special spec can be removed
```

torch_ttnn/passes/lowering/add_data_move_pass.py
jmalone-tt: Interesting, it looks like `_change_layout` is only used for the `AlignSpecInTtnn` spec type, which can have device set to one of `None`, `'host'`, or `TtnnDevice`. That's why I assumed `None` meant we didn't care. If it actually means we need the tensor to *not* be on device, I might change that spec type too (remove the `None` option) or change how things are being handled (explicitly set to `host`). I will experiment further, but if you have any further insights I would appreciate it @kevinwuTT

kevinwuTT: @jmalone-tt Sorry, you're right. The definition has those three types defined `device: Union[None, Type[TtnnDevice], Literal["host"]]` And judging from the code, `None` is used to indicate that we don't want to alter the existing device location of the current node/tensor. So don't care should be the correct interpretation.

jmalone-tt:
Updated to address feedback and clean things up some more.
Remaining questions:
* @ayerofieiev-tt Is there a certain place you are looking to count `to_torch` calls? They should only be inserted when we are going from a ttnn node to a non-ttnn node, but the logic could have issues.
* @kevinwuTT I was able to remove `ttnn.reshape` from the `TTNN_LAYOUT_CHANGE_OPS` list, and renamed the list to be more descriptive. As we eliminate the need to force ROW_MAJOR layouts, we can reduce this set more.

Please let me know if there are any other changes we need before merging

kevinwuTT (approved):

PR #849, opened 2025-03-24
Optimize out conversion from `aten.clone`

kevinwuTT:
### Problem description
`aten.clone` is used often to make sure the tensor is not a view of another tensor which can cause problems for some ops. Pytorch inserts this whenever it needs to. However, when we're dealing with TTNN tensors, we don't need to worry about this, so there's usually no need to convert to `ttnn.clone`. 

### What's changed
Unless `aten.clone` has an input from the graph arguments and is also returned as an output, we will remove the op entirely.

For example, `aten.clone` will not be removed and converted to `ttnn.clone` in this instance only. This is for unit testing purposes.

```
def forward(arg0):
    clone = aten.clone(arg0)
    return clone
```

### DIFF
--- a/tests/lowering/creation/test_clone.py
+++ b/tests/lowering/creation/test_clone.py
@@ -53,6 +53,7 @@ def test_clone_from_arg(device, input_shapes):
     [[(4, 4)]],
 )
 def test_clone_from_node(device, input_shapes):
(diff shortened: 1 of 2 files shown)


jmalone-tt (approved):
Looks good to me, thanks!

PR #851, opened 2025-03-25
Remove additional `ttnn.to_layout` ops

kevinwuTT:
### Problem description
In the past, `ttnn.reshape` had trouble supporting some shapes in `TILE_LAYOUT`, so `ttnn.to_layout` ops are called to mitigate these issues. However, `ttnn.reshape` has improved since, so we can try to remove these workarounds. 

### What's changed
* Remove layout changes for `ttnn.reshape` ops.
* For `ttnn.embedding`, only change layout to `ROW_MAJOR` for the first argument (indices tensor).
* Skip 4 BEIT models and 2 RetinaNet_ResNet50_FPN models. They will be tracked via the following tickets:
  * https://github.com/tenstorrent/pytorch2.0_ttnn/issues/865
  * https://github.com/tenstorrent/pytorch2.0_ttnn/issues/866

### DIFF
--- a/.github/workflows/before_merge.yaml
+++ b/.github/workflows/before_merge.yaml
@@ -257,7 +257,7 @@ jobs:
           fi
           
           python3 tools/send_to_data_team.py \
-              --github_workflow_id ${{ github.event.workflow_run.id }} \
+              --github_workflow_id ${{ github.run_id }} \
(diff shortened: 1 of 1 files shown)


jmalone-tt (approved):
Looks good to me! Happy surprise that there aren't more failing models

ayerofieiev-tt (approved):

PR #859, opened 2025-03-26
Fix workflow id in github workflows

acirikTT:
### Ticket
Not available

### Problem description
`Collect Metrics Report` step of the `collect-metrics` jobs fails due to not using the correct method to get run id. This happens when the before_merge workflow is triggered with anything other than `None`

### What's changed
Uses the correct syntax to get run id


jmalone-tt (approved):
Good catch

PR #868, opened 2025-03-28
Final CI updates for docker container creation and reinstalling dependencies

jmalone-tt:
This PR contains the final steps to get CI stable going forward. It combines the following changes into one PR:
* `before_merge.yaml` is now just a wrapper around the newly created `run-tests.yaml` workflow. The wrapper is needed so we can set a default container for running tests and allow it as an argument in workflow_call flows. This is a limitation of GitHub Actions.
* `run-tests.yaml` defines the core logic for running tests and provides an output that tells whether the tests passed
* `update-docker-container.yaml` is a new workflow that does the following:
    * Check if a docker container already exists given the current dependencies and latest container from tt-metal
    * If needed, builds a new candidate container and pushes it to ghcr
    * Runs test by calling `run-tests.yaml`
    * If tests pass, it will then mark the candidate container with the `latest` tag
* Adds back the step that re-installs dependencies for `run-tests` and `run-accuracy-tests`. The dockerization change removed this step for faster runs, but this caused issues due to updated dependencies not being installed correctly in the container that is used for testing.


jbedichekTT (commented):
Looks good, nice job.

jmalone-tt:
I realized that this will potentially cause errors. Namely, when we run the update dependencies workflow, it will run tests on the old docker container, so it will pass even if updating dependencies will cause issues. Then, the next time we want to run tests, the updated container will have issues. To resolve this, I am updating the PR in the following ways:
* Change update_dependencies workflow to create a candidate container, then run tests in that candidate container, then update the candidate's tag to `latest` only if tests pass. At this point, we are safe to either merge through the merge queue or merge as admin (skipping the queue)
* Add an option to specify the container in the tests workflow, defaulting to the `latest` container
* Keep the update container workflow, but only run it manually (since the expected path is to update dependencies automatically)

With these changes, there is the risk that someone updates dependencies manually and doesn't update the container, but we can add a comment in the file as a reminder

kevinwuTT (approved):

Issue #835, opened 2025-03-20
Investigate low accuracy in CLIP (94.26)


nemanjagrujic:
https://github.com/tenstorrent/pytorch2.0_ttnn/blob/main/docs/models/CLIP/input_variations.md

npetrovic-tenstorrent:
It seems to be failing on ttnn.rsub with the following error 

Traceback (most recent call last):
  File "/home/ubuntu/pytorch2.0_ttnn/tests/autogen_accuracy_tests/CLIP_code.py", line 4152, in <module>
    forward(*inputs)
  File "/home/ubuntu/pytorch2.0_ttnn/tests/autogen_accuracy_tests/CLIP_code.py", line 2626, in forward
    test_accuracy(rsub, ttnn_rsub)
  File "/home/ubuntu/pytorch2.0_ttnn/tests/autogen_accuracy_tests/CLIP_code.py", line 103, in test_accuracy
    assert_with_pcc(expected, actual, pcc = 0.90)
  File "/home/ubuntu/pytorch2.0_ttnn/tests/autogen_accuracy_tests/CLIP_code.py", line 96, in assert_with_pcc
    assert pcc_passed, construct_pcc_assert_message(pcc_message, expected_pytorch_result, actual_pytorch_result)
AssertionError: 0.0

PR #871, opened 2025-03-28
Fix removal of mutations from calling eliminate_dead_code()

kevinwuTT:
### Problem description
`insert_clones_for_input_aliasing` handles the problem with aliasing inputs. It modifies the graph with Torch ops by inserting clones, then removing them after lowering to Aten. (This is before lowering to TTNN). However, this also calls `GraphCleanup()` function which calls `gm.graph.eliminate_dead_code()`. Usually this would be fine in the Aten level because Aten ops do not modify tensors in-place. However, some Torch ops do, and `eliminate_dead_code()` does not handle in-place mutations. For example:

```
1: def f():
2:    mask = torch.full((1, 1), 1)
3:    mask_cond = torch.full((1, 1), True)
4:    masked_fill = mask.masked_fill_(mask_cond, 0)
5:    mask_1 = mask.to(torch.bfloat16)
6:    return mask_1
```
In the function above, `eliminate_dead_code()` does not know that `mask.masked_fill_` mutates the `mask` tensor in-place. Therefore, it sees that `masked_fill` is not used and will eliminate lines 3 and 4. The result of the function above should be `0`. However, after dead code elimination, the function will become the code below and the result will be `1` which is incorrect.

```
1: def f():
2:    mask = torch.full((1, 1), 1)
3:    mask_1 = mask.to(torch.bfloat16)
4:    return mask_1
```

### What's changed
* Remove call to `gm.graph.eliminate_dead_code()` at the Torch level.
* Add unit test to demonstrate mutation
* This also reveals some torch ops that do in-place mutations that were removed previously
  * Fix some issues with `slice_scatter` 
    * Workaround for some 1D concat https://github.com/tenstorrent/tt-metal/issues/20205
  * Turn `copy_` to no-op. 
    * In a scenario where an input to the forward function is supposed to be mutated, handle input aliasing for that input differently
    * Add a unit test for `copy_`
  * test_mobilenet_ssd no longer runs end to end because two aten ops are not converted https://github.com/tenstorrent/pytorch2.0_ttnn/issues/908

## Related Issues:
https://github.com/tenstorrent/pytorch2.0_ttnn/issues/835
This problem was found with the CLIP model after inspecting accuracy discrepancies between the value reported when running from pytest versus the exported standalone accuracy script. This patch should make them consistent.


tests/lowering/misc/test_mutation.py
```diff
@@ -0,0 +1,35 @@
...
+    result_before = m(dummy_input)
+    # The compilation is lazy, so we need to run forward once to trigger the compilation
+
+    m = torch.compile(m, backend=torch_ttnn.backend, options=option)
+    result_after = m(dummy_input)
+    option._out_fx_graphs[0].print_tabular()
+
+    # Check the graph has be rewritten and contain ttnn ops
+    nodes = list(option._out_fx_graphs[0].nodes)
+    # assert [node.target for node in nodes].count(ttnn.add) == 1
```
jmalone-tt, line 32: should this be uncommented?

torch_ttnn/handle_input_aliasing.py
```diff
@@ -27,15 +27,25 @@
...
 # Insert aten.clone nodes after every input to prevent input aliasing
 def insert_clones_for_input_aliasing(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
-    input_nodes = get_input_nodes(gm)
+    # Helper function to determine if a node has in-place mutation. For example, the torch op
+    # `self_tensor.copy_(source_tensor)` will replace the self_tensor with the source_tensor.
+    # The arguments for this op will be (self_tensor, source_tensor) in this order.
+    # TODO: Is there a better way to determine if an op mutates an input without listing manually?
+    def is_node_mutated(node):
+        torch_ops_that_mutate = ["copy_", "torch.slice_scatter"]
+        for user in list(node.users.keys()):
```
jmalone-tt, line 38: a bit confusing that we loop but only ever check the first user (if I'm reading correctly)

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -1149,6 +1149,13 @@ def reshape_1d(code, args=args, kwargs=kwargs):
                 if tensor_shape == src_tensor_shape:
                     return src_tensor
 
+                # Error: Cannot pad RM tensor with specified format if tensors are int
```
jmalone-tt, line 1152: is there a tracking ticket for this error?

torch_ttnn/handle_input_aliasing.py
```diff
@@ -52,7 +62,34 @@ def insert_clones_for_input_aliasing(gm: torch.fx.GraphModule) -> torch.fx.Graph
...
+        the `mask` tensor in-place. Therefore, it sees that `masked_fill` is not used and will eliminate lines 3 and 4. The result of the function above should be `tensor([[0.]], dtype=torch.bfloat16)`. However, after code elimination, the function will become below and
+        the result will be `tensor([[1.]], dtype=torch.bfloat16)` which is incorrect.
+
+        ```
+        1: def f():
+        2:    mask = torch.full((1, 1), 1)
+        3:    mask_1 = mask.to(torch.bfloat16)
+        4:    return mask_1
+        ```
+        """
```
jmalone-tt, line 90: I like the comment here. I don't remember off the top of my head - do we call GraphCleanup elsewhere? I'm wanting to make sure dead code is eliminated at some point

jmalone-tt (changes requested):
Couple of small notes, but looks good overall!

tests/lowering/misc/test_mutation.py
kevinwuTT: Upon further inspection, I don't think this specific test needs to check the converted ops so I removed that line all together. Checking the end result should be enough since if the mutation is removed, regardless of lowering, the final results will mismatch.

torch_ttnn/handle_input_aliasing.py
kevinwuTT: Thanks for catching this. I reworked it to return true if any users in the loop mutates the node which was my original intention.

torch_ttnn/passes/lowering/to_tt_pass.py
kevinwuTT: Thanks for pointing this out. I updated it with a ticket.

torch_ttnn/handle_input_aliasing.py
kevinwuTT: Yeah, looks like we call GraphCleanup in:

- `remove_clones_for_input_aliasing()`
- several places in to_tt_pass

These are called within the `aten_backend()` function so it should be safe at that level and beyond. Calling dead code elimination within `ttnn_backend()` function is not safe from what I've learned.

jmalone-tt (approved):
Thanks for the updates!

PR #892, opened 2025-04-02
Handle nested outputs when calculating accuracy

kevinwuTT:
### Ticket
Issue was discovered because of inconsistent accuracy values between the pytest run and generated accuracy script https://github.com/tenstorrent/pytorch2.0_ttnn/issues/835

### Problem description
Some outputs are nested Dict types. For example, CLIP has an output that has the following keys:

```
CLIPOutput = {
    "logits_per_image",
    "logits_per_text",
    "text_embeds",
    "image_embeds",
    "text_model_output" : {
        "last_hidden_state",
        "pooler_output"
    },
    "vision_model_output" : {
        "last_hidden_state",
        "pooler_output"
    }
}
```
Currently, `calculate_accuracy` silently skips the nested outputs so only the first 4 items are factored into the calculation. 

### What's changed

- Recursively walk through each item including nested Dicts and Lists until torch tensors are found and calculate the mean of the accumulated pcc values.
- Include a unit test to demonstrate the bug
- Compare the whole output of dino model instead just one attribute

### DIFF
--- a/tests/models/dino/test_dino.py
+++ b/tests/models/dino/test_dino.py
@@ -40,6 +40,5 @@ def test_dino(record_property, mode):
 
     tester = ThisTester(model_name, mode)
     results = tester.test_model()
-    results = results.last_hidden_state
 
     record_property("torch_ttnn", (tester, results))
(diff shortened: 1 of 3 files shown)


jmalone-tt (approved):
Looks good to me

ayerofieiev-tt (approved):

Issue #861, opened 2025-03-27
Implement pipeline to run accuracy tests


dgomezTT:
Since the accuracy tests have been removed from the merge pipeline, we need to execute them at least weekly

PR #899, opened 2025-04-04
Fixed accuracy tests

dgomezTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/pytorch2.0_ttnn/issues/861)

### Problem description
Test were not running at all, due some issues in the code generator

### What's changed
Fixed the code generator, that doesn't mean that all the tests are green, it means that now we can actually run them

### DIFF
--- a/tools/generate_op_accuracy_tests.py
+++ b/tools/generate_op_accuracy_tests.py
@@ -388,10 +388,11 @@ def check_accuracy(expected, actual):
     full_input_pkl_path = directory / input_pkl_file
     full_input_pkl_path.parent.mkdir(parents=True, exist_ok=True)
     main_code = f"""
-filepath = Path(__file__).with_name("{input_pkl_file.name}")
-file = lzma.open(filepath, "rb")
-inputs = pickle.load(file)
-forward(*inputs)
+if __name__ == "__main__":
+    filepath = Path(__file__).with_name("{input_pkl_file.name}")
+    file = lzma.open(filepath, "rb")
+    inputs = pickle.load(file)
+    forward(*inputs)
 """
 
     # Assemble all of pieces of code into one script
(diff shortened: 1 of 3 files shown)


tools/generate_op_accuracy_tests.py
```diff
@@ -374,9 +374,9 @@ def _generate_code(model_name, test_accuracy_graph_codes, all_inputs):
         inspect.getsource(assert_with_pcc),
     ]
 
-    # test_accuracy helper function definition
+    # check_accuracy helper function definition
```
jmalone-tt, line 377: Did we have some sort of name collision with `test_accuracy`?

tools/generate_op_accuracy_tests.py
```diff
@@ -388,11 +388,10 @@ def test_accuracy(expected, actual):
     full_input_pkl_path = directory / input_pkl_file
     full_input_pkl_path.parent.mkdir(parents=True, exist_ok=True)
     main_code = f"""
-if __name__ == "__main__":
```
jmalone-tt, line 391: Surprised we need to get rid of this since it's a pretty common pattern. What error was it causing?

jmalone-tt (approved):
Overall looks good! Just left a couple questions for my understanding.

tools/generate_op_accuracy_tests.py
jmalone-tt: nvm, Kevin asked about us using pytest, and I can see why that would cause issues

tools/generate_op_accuracy_tests.py
kevinwuTT: I did a quick test by printing `__name__` and running `pytest` prints out the filename instead of `__main__`. So running `pytest` will end up skipping the entire code in this instance. 😅

ayerofieiev-tt (approved):

tools/generate_op_accuracy_tests.py
dgomezTT: reverted this and changed the code to execute just python

PR #941, opened 2025-04-11
Add support for running models data parallel on mesh devices

jmalone-tt:
### Problem Description
There is an ongoing push for better performance for models in PT2.0. As part of this effort, we are adding support for running models on mesh devices, to better support HW scale out. This PR is the first step - adding support for data parallel execution of models. In particular, it focuses on adding the configuration option for data_parallel and supporting MNIST and BERT models. Other models have not been tested and may not work at this time, but we will gradually add support.

### What's Changed
This PR adds several passes to the compiler:

* `InputAnalysisPass` - analyzes inputs to a model and marks them as parameters, buffers, or arguments. Parameters and buffers will not change during inference, but arguments may be different across batches or forward passes.
* 'MultiDeviceShardAnalysisPass` - marks which nodes will be sharded across mesh devices. This pass only runs if compilation is performed in data_parallel mode, and marks inputs, as well as their transitive users which will be sharded along the batch dimension during execution.
* `MultiDevicePass` - performs the actual sharding to mesh devices and propagates updated shapes throughout the computation graph. Due to limitations imposed by the PyTorch compilation pipeline, this pass just inserts calls to wrappers that mimic the shape changes replicating, sharding, and concatenating tensors will incur. Additionally, this pass will rewrite constant arguments to functions like `aten.view`, `aten._unsafe_view`, and `aten.expand` that reflect the sharded tensors.

This PR also updates existing passes:

* `ToTtPass` - now rewrites the wrapper functions for replicating, sharding, and concatenating tensors to use the corresponding ttnn calls.
* `AddDataMovePass` - handles some cases of concatenating and replicating or sharding tensors during TTNN -> torch or torch -> TTNN transitions. There was also a special case for aligning in TTNN due to layout not being supported on host for MultiHostStorageTensors.

Additionally, there are changes for testing. Now, there are tests for BERT and MNIST in `tests/multi_device` that will be run by CI during the `run-tests` workflow.

### Other Notes

The performance of this implementation is bad. Depending on runs, I have observed a slowdown compared to running on a single device with a batch size of 2. This may be better running at higher batch sizes, but I have not tested this thoroughly. My hope is to merge this initial change that gets data parallel runs working, then improve performance in other PRs, so we don't have a long-running separate branch.

I am not crazy about the special cases in `AddDataMovePass` or the double rewrite between `MultiDevicePass` and `ToTtPass`, but this achieves the desired functionality. Both of these are areas for improvement in the future.


torch_ttnn/passes/multi_device_pass.py
```diff
@@ -0,0 +1,127 @@
...
+from torch.fx.passes.infra.pass_base import PassBase, PassResult
+from torch_ttnn.passes.lowering import target_wrappers
+from torch_ttnn.passes.analysis.input_analysis_pass import PrimalTag
+
+import math
+
+import ttnn
+
+
+class GraphWrapper:
```
ayerofieiev-tt, line 15: please add a short docstring

torch_ttnn/passes/multi_device_pass.py
```diff
@@ -0,0 +1,127 @@
...
+                    elif node.meta["primal_tag"] == PrimalTag.ARGUMENT:
+                        self.shard_to_mesh(wrapper, node)
+                except Exception as e:
+                    print(f"Warning: Could not distribute node {node.name}: {e}")
+
+            elif node.op == "output":
+                try:
+                    self.concat_to_tensor(wrapper, node)
+                except Exception as e:
+                    print(f"Warning: Could not concat node {node.name}: {e}")
```
ayerofieiev-tt, line 110: made me think we should replace prints with logger

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -553,6 +554,46 @@ def reshape_1d(code, args=args, kwargs=kwargs):
                     node_user.replace_all_uses_with(new_node)
                 return None
 
+            if node.target == target_wrappers.replicate_tensor:
+                spec_dtype = TtnnBfloat16()
+                if get_dtype(node.args[0]) in [torch.int32, torch.int64]:
+                    spec_dtype = TtnnUint32()
```
ayerofieiev-tt, line 560: what if none of those?

torch_ttnn/passes/lowering/target_wrappers.py
```diff
@@ -21,8 +23,24 @@ def pack_to_tuple(*args):
...
 
+@torch.fx.wrap
+def from_device(device_tensor):
+    if device_tensor.storage_type() == ttnn.StorageType.MULTI_DEVICE:
+        device_tensor = ttnn.get_device_tensors(device_tensor)[0]
+    return ttnn.from_device(device_tensor)
+
+
+@torch.fx.wrap
+def to_device(device_tensor, device=None):
```
ayerofieiev-tt, line 34: what does it mean when this is is a `device_tensor` and we call `to_device`?

torch_ttnn/passes/lowering/target_wrappers.py
```diff
@@ -21,8 +23,24 @@ def pack_to_tuple(*args):
     return tuple(args)
 
 
+@torch.fx.wrap
+def from_device(device_tensor):
+    if device_tensor.storage_type() == ttnn.StorageType.MULTI_DEVICE:
+        device_tensor = ttnn.get_device_tensors(device_tensor)[0]
+    return ttnn.from_device(device_tensor)
```
ayerofieiev-tt, line 30: I'd expect we need to collect (join responses from each device) here, or its not the case?

.github/actions/common_multi_device_tests/action.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Run Multi Device Tests
+      shell: bash
+      run: |
+        set +e
+        if [ "${{ github.event.inputs.commit_report }}" == "None" ]; then
+          num_iterations=1
+        else
+          num_iterations=5
+        fi
+        python3 -m pytest --github-report tests/multi_device/ --data_parallel --report_nth_iteration=$num_iterations --gen_op_accuracy_tests -s
```
kevinwuTT, line 20: Just a note: Depending on which commit gets merged first, `--gen_op_accuracy_tests` will be changed to `--export_code=accuracy`.

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -392,11 +393,22 @@ def _change_layout(self, spec):
 
         input_node = spec.input_node
 
-        if need_from_device:
-            input_node = self.graph.call_function(ttnn.from_device, (input_node,))
+        # TODO: always layout on device once that is supported (once tilize uint32 works on device)
```
ayerofieiev-tt, line 396: https://github.com/tenstorrent/tt-metal/pull/20114

tests/multi_device/test_multi_mnist.py
kevinwuTT, line 1: I might be overlooking, but is there anything new with this test?

torch_ttnn/passes/analysis/multi_device_shard_analysis_pass.py
```diff
@@ -0,0 +1,57 @@
...
+            modified = False
+            return PassResult(gm, modified)
+
+        node_list = list(gm.graph.nodes)
+        for node in node_list:
+            if node.op == "placeholder" and node.meta.get("primal_tag") == PrimalTag.ARGUMENT:
+                shard_dim = 0
+                concat_size = node.meta["val"].shape[shard_dim]
+
+                # TODO: remove limitation that sharded dimension size is multiple of batch size
```
ayerofieiev-tt, line 48: this comment is a bit confusing. Do you say something like Tensor Size should be divisible by # of devices in the mesh?

torch_ttnn/passes/analysis/multi_device_shard_analysis_pass.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class MultiDeviceShardAnalysisPass(PassBase):
+    def __init__(self, device):
+        super().__init__()
+
+        self.device = device
+
+    def call(self, gm: torch.fx.GraphModule):
+        if not isinstance(self.device, ttnn._ttnn.multi_device.MeshDevice):
```
ayerofieiev-tt, line 38: > ttnn._ttnn.multi_device.MeshDevice
Leaking this is not good.

But FYI, this is changing. 
https://github.com/tenstorrent/tt-metal/pull/18067

Ask Stas for details about "unit mesh"

ayerofieiev-tt: Basically all single device cases converge into a mesh of size 1

torch_ttnn/passes/analysis/input_analysis_pass.py
```diff
@@ -0,0 +1,41 @@
...
+    PARAMETER = 1
+    BUFFER = 2
+    ARGUMENT = 3
+
+
+class InputAnalysisPass(PassBase):
+    def __init__(self, n_parameters, n_buffers, n_arguments):
+        super().__init__()
+
+        self.n_parameters = n_parameters
```
ayerofieiev-tt, line 19: do I read `n_parameters` as `parameters count`?

torch_ttnn/passes/analysis/input_analysis_pass.py
```diff
@@ -0,0 +1,41 @@
...
+        self.n_buffers = n_buffers
+        self.n_arguments = n_arguments
+
+    def call(self, gm: torch.fx.GraphModule):
+        def annotate_node(node, tag):
+            node.meta["primal_tag"] = tag
+
+        modified = False
+        placeholder_counter = 0
```
ayerofieiev-tt, line 28: would like to learn more about the approach here.
I don't understand how this works

torch_ttnn/passes/analysis/input_analysis_pass.py
```diff
@@ -0,0 +1,41 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import torch
+from torch.fx.passes.infra.pass_base import PassBase, PassResult
+from enum import Enum
+
+
+class PrimalTag(Enum):
```
ayerofieiev-tt, line 9: please leave a docstring explaining the difference

e.g. it is not really clear what buffer is. 

and I assume `parameter` is model parameter like weights and biases, while `argument` is like an `input` (just give another example of an argument thats not a user input)

torch_ttnn/backend.py
```diff
@@ -132,9 +139,25 @@ def aten_backend(
     # Register ttnn objects as graph globals
     register_ttnn_objects(option)
 
-    # Rewrite with ttnn ops, will insert redundant data movement
+    # Run analysis passes to help with ttnn ops
     from torch.fx.passes.infra.pass_manager import PassManager
+    from torch_ttnn.passes.analysis.input_analysis_pass import InputAnalysisPass
+    from torch_ttnn.passes.analysis.multi_device_shard_analysis_pass import MultiDeviceShardAnalysisPass
+
+    passes = [
```
ayerofieiev-tt, line 147: why can't these be merged with the main list of passes?

tests/multi_device/test_multi_mnist.py
```diff
@@ -0,0 +1,68 @@
...
+from torchvision import transforms, datasets
+from torch.utils.data import DataLoader
+
+import torch.nn as nn
+import torch.nn.functional as F
+from tests.utils import ModelTester
+
+
+# adapted from https://github.com/pytorch/examples/blob/main/mnist/main.py
+class MnistModel(torch.nn.Module):
```
ayerofieiev-tt, line 15: would be really interesting to test what backend decided params, buffers, args are.
would also be awesome to inspect the graph before and after transformations, maybe even visually

ayerofieiev-tt: same for bert

torch_ttnn/passes/multi_device_pass.py
```diff
@@ -0,0 +1,127 @@
...
+            if node.op == "placeholder":
+                try:
+                    if (node.meta["primal_tag"] == PrimalTag.PARAMETER) or (
+                        node.meta["primal_tag"] == PrimalTag.BUFFER
+                    ):
+                        self.replicate_to_mesh(wrapper, node)
+                    elif node.meta["primal_tag"] == PrimalTag.ARGUMENT:
+                        self.shard_to_mesh(wrapper, node)
+                except Exception as e:
+                    print(f"Warning: Could not distribute node {node.name}: {e}")
```
kevinwuTT, line 104: Should we use `logging.info` or `logging.warning` instead of print?

tests/conftest.py
```diff
@@ -60,20 +61,33 @@ def input_var_check_ttnn(request):
...
+def device(request):
     # TODO(tt-metal#13746): Currently L1 small size needs to be manually determined
-    device_id = 0
     l1_small_size = 16384
     dispatch_core_config = get_dispatch_core_config()
 
-    device = ttnn.open_device(device_id=device_id, dispatch_core_config=dispatch_core_config, l1_small_size=16384)
+    if request.config.getoption("--data_parallel"):
+        device = ttnn.open_mesh_device(
+            ttnn.MeshShape(1, 2), dispatch_core_config=dispatch_core_config, l1_small_size=l1_small_size
```
ayerofieiev-tt, line 71: we can later allow user to specify this info

ayerofieiev-tt: I will request us a t3k machine to enable larger scale tests

ayerofieiev-tt (approved):

kevinwuTT (approved):
Looks good! Just made a few questions.

torch_ttnn/passes/lowering/to_tt_pass.py
jmalone-tt: If not int32 or int64, the line above defines type as Bfloat16. Changed this to use an `else` so it's clearer though

torch_ttnn/passes/lowering/target_wrappers.py
jmalone-tt: Poorly named variable, we expect this to be a host_tensor (renamed). It looks like calling to_device on a tensor that is already on device doesn't cause problems though.

torch_ttnn/passes/lowering/target_wrappers.py
jmalone-tt: Thanks for the comment here, after looking more in-depth, these functions are unused. Must have been a hack from bringup we no longer need. Deleted

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -408,14 +420,29 @@ def _create_aligned_node(self, spec):
             kwargs = {}
             if spec.device is not None and spec.device != "host":
                 kwargs["device"] = spec.device()
+                if spec.input_node.meta.get("is_sharded"):
```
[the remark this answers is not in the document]
jmalone-tt: Agreed. This implementation "infects" MultiDevicePass, ToTtPass, and AddDataMovePass. I will add comments to the files that are not multi device agnostic, and try to reduce them in the future

torch_ttnn/passes/analysis/multi_device_shard_analysis_pass.py
jmalone-tt: Updated. Your comment is clearer, but we are particularly interested in the sharding dimension here.

torch_ttnn/passes/analysis/input_analysis_pass.py
jmalone-tt: Yes. Changed to `num_*` so it's clear

torch_ttnn/passes/analysis/input_analysis_pass.py
jmalone-tt: Added a comment to explain for the call function. This relies on how PyTorch lays out arguments under the hood, which is also relied on elsewhere in the stack (I don't remember off the top of my head)

tests/multi_device/test_multi_mnist.py
jmalone-tt: When we print args, it shows the primal_tag. From what I remember, MNIST has all parameters except 1 argument, BERT has 3 arguments (don't remember parameters / buffers off the top of my head). 

Agreed it would be helpful to have some more visualization in the compiler 🤔

tests/conftest.py
jmalone-tt: I have been thinking about this. One simple way would be based on device shape passed in. We could say one dimension is data parallel split, and the other is tensor parallel split. For example, MeshShape(4,2) could be running model with data parallelism x4 and tensor parallelism x2. This might not give enough control, but seems pretty easy to configure for users. Do you have any thoughts?

tests/multi_device/test_multi_mnist.py
jmalone-tt: The main differences are increasing batch_size, not running on train, and not marking eval end to end. Otherwise it is very similar to tests/models/mnist/test_mnist

If you have better ideas on how to test multi-device, I am definitely open to suggestions. I'm not crazy about the duplication here, but didn't have a good way to change pytest marks or batches without impacting the normal model tests

kevinwuTT: Gotcha. Yeah I think this fine as is. Sometimes it's clearer to just have a separate test vs modifying an existing one. Since you added a pytest option to enable multi device, you can technically pass in a fixture called `pytestconfig` [(reference)](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.fixtures.pytestconfig) into the existing Mnist test, and add conditions around that. I think you can also control the end-to-end pytest mark too with extra fixtures, but this might end up making the test messy.

PR #961, opened 2025-04-16
Enables program cache for testing models

jmalone-tt:
### Problem description
It recently came up that we test models without program cache enabled, which hurts the performance we show in the README. This change enables program cache when testing.

Based on a quick check, this negatively impacts the first run time (since we are creating the cache). However, it positively impacts the 5th run time (MNIST went to ~10ms, a 3x speedup; BERT went to ~2600ms, a 10% speedup)

There are some ops that fail with program cache enabled due to OOM errors. In total, this PR removes 5 end-to-end models and 11 compiled models.

### What's changed
Added lines to enable program cache
Added lines to reset program cache between tests to work around issue with eltwise ops ([ticket here](https://github.com/tenstorrent/tt-metal/issues/21086#issuecomment-2827130662))
Marked newly failing models as xfail with explaining reason


jmalone-tt:
Need to run tests to see if program cache breaks anything. If adding cache does break tests, we might need to be more strategic about how we add it.

jmalone-tt:
Will update description, but by my count, we lose 5 E2E models and 11 compiled models with this change. Still need to reduce reproduction steps and fire tickets

ayerofieiev-tt (approved):

PR #970, opened 2025-04-22
Support <2 rank in Tile layout

jmalone-tt:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/pytorch2.0_ttnn/issues/442)

### Problem description
There are several places in ToTtPass where we guard against scalar inputs for functions that now have support from the tt-metal side. By removing these guards and updating tests, we can convert more calls to ttnn calls.

### What's changed
This PR has separates commits for each part of the ToTtPass that has been updated. Each commit also updates relevant tests to exercise the new behavior.

Note that this change leaves one reference to issue 422 in ToTtPass, but it is resolved by https://github.com/tenstorrent/pytorch2.0_ttnn/pull/968

### DIFF
--- a/tests/lowering/eltwise/binary/test_add.py
+++ b/tests/lowering/eltwise/binary/test_add.py
@@ -21,6 +21,7 @@ class AddModule(torch.nn.Module):
         ((32, 32), (32, 32)),
         ((64,), (32, 64)),
         ((64, 32), (64, 1)),
+        ((64, 32), ()),
         pytest.param(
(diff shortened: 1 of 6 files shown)


kevinwuTT (approved):

ayerofieiev-tt:
Please improve the PR title, it becomes the commit message once fixed and this “issue 442” should be rather replaced with an actual brief

ayerofieiev-tt (approved):

PR #973, opened 2025-04-23
Lower aten.view and aten._unsafe_view to ttnn.experimental.view when valid

jmalone-tt:
### Problem description
While working on performance tuning for BERT and multi-device BERT, we realized there are many (450+) calls to ttnn.reshape in the generated code. Nearly all of these are due to the lowering of `aten.view` and `aten._unsafe_view`. This PR lowers to ttnn.experimental.view when valid. The benefit is that ttnn.experimental.view is guaranteed to be zero copy, so it is easier to understand the performance implications just by reading generated code.

### What's changed
Updates ToTtPass to lower `aten.view` to `ttnn.experimental.view`
Updates tests

### DIFF
--- a/torch_ttnn/passes/lowering/to_tt_pass.py
+++ b/torch_ttnn/passes/lowering/to_tt_pass.py
@@ -897,8 +897,8 @@ def ReplaceMoreTtManually(gm: torch.fx.GraphModule, use_less_ttnn_op_types: bool
                 second_last_dimension_unpadded = (
                     len(input_shape) > 1
                     and len(output_shape) > 1
-                    and input_shape[-2] % 32 == 0
-                    and output_shape[-2] % 32 == 0
+                    and input_shape[-2] % ttnn.TILE_SIZE == 0
+                    and output_shape[-2] % ttnn.TILE_SIZE == 0
                 )
 
                 if last_dimension_constant and (second_last_dimension_constant or second_last_dimension_unpadded):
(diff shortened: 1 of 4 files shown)


torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -878,7 +878,30 @@ def reshape_1d(code, args=args, kwargs=kwargs):
...
+                # Be cautious and assume Layout::TILE for now
+
+                last_dimension_constant = input_shape[-1] == output_shape[-1]
+                second_last_dimension_constant = (
+                    len(input_shape) > 1 and len(output_shape) > 1 and input_shape[-2] == output_shape[-2]
+                )
+                second_last_dimension_unpadded = (
+                    len(input_shape) > 1
+                    and len(output_shape) > 1
+                    and input_shape[-2] % 32 == 0
```
ayerofieiev-tt, line 897: Ideally instead of %32 we should get tile size from the tensor spec

jmalone-tt: I think this is fixed now, but if there's a better way, please let me know

ayerofieiev-tt: This is better, but not perfect.
Lets leave as is for now.

Ideal (not needed now) is to read TensorSpec from Tensor and read Tile size from Tensor Spec

dgomezTT (approved):
looks good to me! I'll keep this PR around as reference to study the compiler a bit further

PR #979, opened 2025-04-25
Additional small fixes for exporting code

kevinwuTT:
### Problem description
1. Some aten ops use string literals but the strings are not quoted `aten._assert_async.msg(scalar_tensor, "assertion error")`. 
1. Some Torch tensors with scalars and constants are not converted correctly. e.g. `torch.[0]` instead of `torch.tensor([0])`
1. Filenames with spaces are not found when using the CI
1. When using [graph tracing](https://github.com/tenstorrent/tt-metal/blob/main/tech_reports/ttnn/graph-tracing.md) some of the tensors differ in the exported code vs compiled run. 

### What's changed
1. For some ops with string literal arguments, the argument type is checked if it is an instance of a `str`. Add quotes around that argument if so. This was observed in the `stable_diffusion/test_stable_diffusion_v2.py`
1. Calling `str()` around a tensor object would include the `tensor(` type in the string. Then prepend `torch.`.
1. Add quotes around the filename for the Accuracy Test CI action.
1. Match the `dispatch_code_config` of the device in the exported code.
1. Match mesh device initialization in the exported code


tools/export_code.py
```diff
@@ -537,8 +559,10 @@ def format_forward_calls(call_forwards_in_main, indents=""):
     filepath = Path(__file__).with_name("{input_pkl_file.name}")
     file = lzma.open(filepath, "rb")
     inputs = pickle.load(file)
-    device = ttnn.open_device(device_id=0, l1_small_size=16384)
+    device = ttnn.open_device(device_id=0, dispatch_core_config=get_dispatch_core_config(), l1_small_size=16384)
```
jmalone-tt, line 562: I'm not as familiar with the rest of this file. Is there the ability to follow conftest.py here and insert calls to open_mesh_device, etc. if that is what is used for the test?

kevinwuTT: @jmalone-tt I pushed an update to support opening mesh devices in the exported code. In the future, I would like to see if there's a clean way to dynamically copy the device initialization from conftest.py. Using `inspect.getsource` for the device fixture isn't as straightforward because of the `request` pytest type. I don't want the standalone code to depend on pytest if possible.

tools/export_code.py
```diff
@@ -550,6 +554,22 @@ def format_forward_calls(call_forwards_in_main, indents=""):
...
+            "device = ttnn.open_mesh_device(ttnn.MeshShape(1, 2), dispatch_core_config=dispatch_core_config, l1_small_size=l1_small_size)"
+        ]
+        close_device = "ttnn.close_mesh_device(device)"
+    else:
+        open_device = [
+            "device = ttnn.open_device(device_id=0, dispatch_core_config=dispatch_core_config, l1_small_size=l1_small_size)",
+            "ttnn.SetDefaultDevice(device)",
+        ]
+        close_device = "ttnn.close_device(device)"
+    open_device = "\n".join(f"{_get_indent(1)}{i}" for i in open_device)
```
jmalone-tt, line 639: Might also be nice to add `device.enable_program_cache()` to `open_device` list. Not needed, so it's up to you if we include that here or in a different PR some other time

jmalone-tt (approved):
Thanks for the update! Left one comment about enabling program cache. Agreed that it would be good if we could avoid the copy-paste of what's in conftest, but I think this is good to merge

tools/export_code.py
```diff
@@ -552,10 +570,18 @@ def _save_to_disk(model_name, forward_codes, call_forwards_in_main, all_inputs,
...
+        inspect.getsource(get_dispatch_core_type),
+        inspect.getsource(get_dispatch_core_axis),
+        inspect.getsource(get_dispatch_core_config),
+    ]
+
     # check_accuracy helper function definition
     check_accuracy_code = (
         """
 def check_accuracy(expected, actual):
+    if expected == None and actual == None:
```
philei-tt, line 583: I'm sorry for nitpicking, but it's recommended to compare with "is" instead of "==" when comparing to None :)

tools/export_code.py
```diff
@@ -597,6 +623,22 @@ def format_forward_calls(call_forwards_in_main, indents=""):
     else:
         forward_calls_joined = format_forward_calls(call_forwards_in_main, _get_indent(1))
 
+    # TODO: Find a more dynamic way to copy device initialization from conftest.py
+    if torch_ttnn_option.data_parallel:
+        # Support mesh device
+        open_device = [
+            "device = ttnn.open_mesh_device(ttnn.MeshShape(1, 2), dispatch_core_config=dispatch_core_config, l1_small_size=l1_small_size)"
```
philei-tt, line 630: I don't quite understand everything in this code, but is it ok to hardcode mesh shape here?

tools/export_code.py
kevinwuTT: Done. Not sure why I used "==" when I used "is" in other parts of the code. Thanks for catching this.

tools/export_code.py
kevinwuTT: This part is a copy of the device initialization that happens in https://github.com/tenstorrent/pytorch2.0_ttnn/blob/23046d222e962a5e5486dd0397e0d1bc233c90bb/tests/conftest.py#L68. In the future, it would be nice to figure out a cleaner way without manually hardcoding any lines. 

In regards to the MeshShape, it's taken from https://github.com/tenstorrent/pytorch2.0_ttnn/blob/23046d222e962a5e5486dd0397e0d1bc233c90bb/tests/conftest.py#L76 Joe said that it will be user configurable in the future, but for now we've only tested on n300s with that configuration.

PR #990, opened 2025-04-30
Update TT-NN to 0.58.0-rc15

ayerofieiev-tt:
This PR updates TT-NN wheel to the latest pre-release version.


jmalone-tt:
I believe this fail is due to recent metal changes for concat. Created a ticket to track bug here:
https://github.com/tenstorrent/tt-metal/issues/21494

PR #991, opened 2025-04-30
Fusing operations that can be represented by a higher level op

dgomezTT:
### Ticket
[Link to Github Issue
](https://github.com/tenstorrent/pytorch2.0_ttnn/issues/992)

### Problem description
There are some operations in the compiled code that can be unified in a higher level call. A good example of this is linear and attention_soft_max

### What's changed
Added a new pass that will take care of unifying common patterns into a high level call. This should reduce the amount of operations, the amount of interactions with python and hopefully the amount of reshapes and data handling

Before PR:
Runtime Iteration without cache: 9.11285948753357
Runtime Iteration with cache: 0.38252902030944824

After PR:
Runtime Iteration without cache: 8.339576005935669
Runtime Iteration with cache: 0.37560153007507324

Generated code here:
https://github.com/dgomezTT/BertCompiledOptimization/commit/1b31451af6c13ffaf66816bea95524afe67baad3

Hardware used:
n150d + ryzen 9 5950x3d


dgomezTT:
https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/14781806185

torch_ttnn/backend.py
```diff
@@ -163,6 +164,7 @@ def aten_backend(
         EliminateCoreopsPass(),
         CSEPass(),
         PermuteReshapeTuple(),
+        FusionPass(),
```
kevinwuTT, line 167: Are we required to look for to_torch/from_torch ops? If not, can we put this before the AddDataMovePass to simplify the pattern matching?

dgomezTT: that is a great suggestion! I'll get right on it

torch_ttnn/passes/patterns/linear_pattern.py
```diff
@@ -0,0 +1,118 @@
...
+                if activation[0] is None:
+                    potential_replaceable_view = None
+                matches.append((transpose, matmul, add, potential_replaceable_view, activation))
+
+        return matches
+
+    def replace_linear(self, matches: List[Tuple[torch.fx.Node, ...]]):
+        for match in matches:
+            if match == matches[-1]:
+                print("last match")
```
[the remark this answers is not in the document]
dgomezTT: Right on it! It was something I was using to troubleshoot a test

dgomezTT:
https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/14801021798

torch_ttnn/backend.py
dgomezTT: addressed!

torch_ttnn/passes/patterns/linear_pattern.py
```diff
@@ -0,0 +1,127 @@
...
+                if gelu:
+                    activation = ("gelu", gelu)
+                elif relu:
+                    activation = ("relu", relu)
+                elif silu:
+                    activation = ("silu", silu)
+
+                if activation[0] is None:
+                    potential_replaceable_view = None
+                matches.append((transpose, matmul, add, potential_replaceable_view, activation))
```
ayerofieiev-tt, line 83: maybe something like
```py
# Create a mapping of activation function types to their string names
activation_map = {
    ttnn.gelu: "gelu",
    ttnn.relu: "relu",
    ttnn.silu: "silu"
}

potential_replaceable_view = self._find_exclusive_user_of_type(add, ttnn.experimental.view)
if potential_replaceable_view is not None:
    # Default activation to None
    activation = (None, None)
    
    # Check each activation type from the map
    for act_func, act_name in activation_map.items():
        act_node = self._find_single_user_of_type(potential_replaceable_view, act_func)
        if act_node:
            activation = (act_name, act_node)
            break
            
    # If no activation was found, set view to None
    if activation[0] is None:
        potential_replaceable_view = None
        
    matches.append((transpose, matmul, add, potential_replaceable_view, activation))
```

torch_ttnn/passes/patterns/soft_max_pattern.py
```diff
@@ -0,0 +1,91 @@
...
+        # (int, float bigger than 0)
+        multiply_nodes = self._find_nodes_of_type(ttnn.multiply)
+        for multiply in multiply_nodes:
+            # Check for the scale factor (1/sqrt(head_size))
+            # this number should be always be positive
+            if not (
+                len(multiply.args) > 1
+                and isinstance(multiply.args[1], (int, float))
+                and multiply.args[1] > 0
+                and multiply.args[1] < 1
```
ayerofieiev-tt, line 37: what is the `int` value that can satisfy this range?

ayerofieiev-tt (approved):

torch_ttnn/passes/patterns/soft_max_pattern.py
dgomezTT: it should be only float, I'll remove the int

torch_ttnn/passes/patterns/linear_pattern.py
```diff
@@ -0,0 +1,127 @@
...
+
+        return matches
+
+    def replace_linear(self, matches: List[Tuple[torch.fx.Node, ...]]):
+        for match in matches:
+            transpose, matmul, add, view, activation = match
+
+            bias_arg = add.args[1] if add.args[0] is matmul else add.args[0]
+            # If there is no activation, keep the last view node
+            with self.gm.graph.inserting_after(activation[1] if activation[0] else add):
```
jmalone-tt, line 93: Might be easier to read if we factor out the `activation[1] if activation[0] else add` into a variable. Can use it several places then too instead of repeating the same logic

torch_ttnn/passes/patterns/pattern_matcher_base.py
```diff
@@ -0,0 +1,101 @@
...
+import torch
+import torch._dynamo
+from typing import Optional, Tuple, List, TypeVar, Generic
+
+# Type variable for pattern matching results
+# Can be a tuple of nodes, a single node, or any other pattern structure
+PatternType = TypeVar("PatternType")
+
+
+class PatternMatcherBase(Generic[PatternType]):
```
jmalone-tt, line 14: Are there any transforms you have planned that will not return `Tuple[torch.fx.Node, ...]`? If so, this is good to keep, otherwise it might be a premature abstraction

torch_ttnn/passes/patterns/pattern_matcher_base.py
```diff
@@ -0,0 +1,101 @@
...
+        Remove nodes from the graph. Since ttnn.to_torch checks are handled during pattern matching,
+        we can safely remove these nodes.
+
+        Args:
+            nodes: List of nodes to be removed (can contain None values which will be skipped)
+        """
+
+        for node in nodes:
+            if node is not None:
+                node.users.clear()  # Clear users to ensure clean removal
```
jmalone-tt, line 100: I think other places we just call `erase_node`. What is the failure case if we don't call `node.users.clear()`?

torch_ttnn/passes/patterns/linear_pattern.py
```diff
@@ -0,0 +1,127 @@
...
+                matmul,
+                transpose,
+            ]
+            self.safe_remove_nodes(nodes_to_remove)
+
+        self.gm.graph.lint()
+        self.gm.recompile()
+        return matches
+
+    def match_pattern(self) -> List[Tuple[torch.fx.Node, ...]]:
```
jmalone-tt, line 121: Is there a reason `match_pattern` doesn't just have the logic from `self.match_linear_pattern` instead of making that a separate function? Same with `replace_pattern` / `self.replace_linear` and equivalents in `softmax_pattern.py`

torch_ttnn/passes/patterns/soft_max_pattern.py
```diff
@@ -0,0 +1,91 @@
...
+    def match_softmax_pattern(self) -> List[Tuple[torch.fx.Node, ...]]:
+        """
+        Match the attention softmax pattern:
+        multiply(x) -> add(attention_mask) -> softmax(numeric_stable=True)
+
+        This pattern represents:
+        1. Division by sqrt(head_size)
+        2. Addition of attention mask
+        3. Numerically stable softmax computation
+        """
```
jmalone-tt, line 24: I like the docstrings here - makes it easy to understand what we are matching. Maybe show what it will be replaced with too?

jmalone-tt (commented):
Thanks for getting all of this working! Pumped to have some op fusion in the compiler!

kevinwuTT (approved):
Thanks for addressing the pass placement order. LGTM

torch_ttnn/passes/patterns/pattern_matcher_base.py
dgomezTT: this is mostly me dealing with my obsession with typing everything, I will get rid of it

torch_ttnn/passes/patterns/pattern_matcher_base.py
dgomezTT: this was mostly a different problem that I solved by being way more specific in the nodes I care to transform. I can delete this

jmalone-tt (approved):
LGTM!

PR #996, opened 2025-05-02
Improve data parallel performance

jmalone-tt:
This PR primarily improves DP performance by reducing unnecessary data movement due to calling from_torch with suboptimal tensor specs. This incurred additional data movement when the AddDataMovePass was later run.

This PR also reduces the number of passes that must be "aware" of whether we are running with data parallelism by removing checks from the ToTtPass.

Finally, the sequence length is changed for BERT tests to better compare against tt-metal performance numbers and some checks for whether the compiler is running data parallel are also fixed.


torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -500,7 +531,7 @@ def _connect_aligned_node_layer_norm(
         else:
             self._connect_aligned_node(node, aligned_node, input_site, input_site_type)
 
-    def align(self, node, input_node, input_site, input_site_type: InputSiteType):
+    def align(self, node, input_node, input_site, input_site_type: InputSiteType, first_node):
```
jmalone-tt, line 534: The changes below here in this file move all the `from_torch` calls to the top of the function. This makes it easier to manually hoist loading weights to another function. Happy to get rid of it in case anyone thinks it's ugly though. I don't believe there's a perf impact

tests/models/bert/test_bert.py
```diff
@@ -19,14 +19,49 @@ def _load_model(self):
...
     def _load_inputs(self):
         # Set up sample input
-        self.context = 'Johann Joachim Winckelmann was a German art historian and archaeologist. He was a pioneering Hellenist who first articulated the difference between Greek, Greco-Roman and Roman art. "The prophet and founding hero of modern archaeology", Winckelmann was one of the founders of scientific archaeology and first applied the categories of style on a large, systematic basis to the history of art. '
-        self.question = "What discipline did Winkelmann create?"
-        inputs = self.tokenizer.encode_plus(
-            self.question,
-            self.context,
+        batch_size = 1
+        self.context = [
+            'Johann Joachim Winckelmann was a German art historian and archaeologist. He was a pioneering Hellenist who first articulated the difference between Greek, Greco-Roman and Roman art. "The prophet and founding hero of modern archaeology", Winckelmann was one of the founders of scientific archaeology and first applied the categories of style on a large, systematic basis to the history of art. ',
```
dgomezTT, line 24: Can we do this another tests? So we have a test with batch 1 and batch 8? I did this locally before but seems it will be useful for other people

dgomezTT (approved):
Looks good to me and something we really need to have! Btw, maybe is worth adding the metrics to compare how this improved the situation?

ayerofieiev-tt (commented):
would be good to move out data to json outside of .py

torch_ttnn/passes/lowering/add_data_move_pass.py
philei-tt: I think it might increase memory consumption thought. In theory you can deallocate tensor used earlier in graph before some other is even moved to device.

kevinwuTT: Yeah that makes sense, since there is a higher chance some inputs won't be used until later in the graph, but they will still occupy memory until then. But at the same time, I think caching weights via native device has a similar memory usage pattern. So in my observation this change is like a precursor to that.

philei-tt: But you want to keep weights in the memory all the time anyway, since you will need them on backpropagation + next iteration. Can't say that about inputs though

torch_ttnn/backend.py
```diff
@@ -184,8 +185,7 @@ def aten_backend(
     pm = PassManager(passes=passes)
     gm, modified = pm(gm)
 
-    gm.graph.lint()
-    gm.recompile()
+    GraphCleanup(gm)
```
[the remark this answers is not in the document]
jmalone-tt: That's a good point. I've been looking into marking nodes impure so they aren't removed by DCE. Looks like we can mark an arbitrary function call as impure with `torch.fx.node.has_side_effect`. I'll wait until your change is merged then add support for that!

tests/models/bert/test_bert.py
jmalone-tt: I agree this is generally useful. I think there's an open PR that implements a lot of this, so I will address it separately

torch_ttnn/passes/lowering/add_data_move_pass.py
jmalone-tt: That's true. It hasn't shown up as an issue yet causing OOMs, but we should be able to keep inputs just before the nodes that use them by using the tags from an earlier analysis pass. Thanks for catching that

torch_ttnn/backend.py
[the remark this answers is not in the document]
kevinwuTT: Nice finding! Looks good!

kevinwuTT (approved):

PR #1002, opened 2025-05-05
Add a pass to explicitly deallocate ttnn tensors after last use

kevinwuTT:
### Problem description
Although the previous PR https://github.com/tenstorrent/pytorch2.0_ttnn/pull/980 has improved the memory allocation for the exported code, that implementation depends on the Python garbage collector. In the case of TTNN Tensors, that is not deterministic. One way to test this is to use an indirect reference. For example, in the code below, Pytorch fx.compile will insert `tensor = None` code whenever it determines the last-use of a tensor. However, the two TTNN tensors were never really deallocated because the data is still valid when calling `ttnn.to_torch` at the end.

```
import torch
import ttnn

torch_tensor_0 = torch.rand((32,32), dtype = torch.bfloat16)
torch_tensor_1 = torch.rand((32,32), dtype = torch.bfloat16)

with ttnn.manage_device(device_id=0) as device:
    ttnn_tensor_0 = ttnn.from_torch(torch_tensor_0, layout = ttnn.TILE_LAYOUT, device = device)
    ttnn_tensor_1 = ttnn.from_torch(torch_tensor_1, layout = ttnn.TILE_LAYOUT, device = device)
    pack_to_tuple = tuple([ttnn_tensor_0, ttnn_tensor_1]); ttnn_tensor_0 = ttnn_tensor_1 = None
    getitem_0 = pack_to_tuple[0]
    getitem_1 = pack_to_tuple[1]; pack_to_tuple = None
    to_torch_0 = ttnn.to_torch(getitem_0); getitem_0 = None
    to_torch_1 = ttnn.to_torch(getitem_1); getitem_1 = None

```

However, if you replace the `ttnn_tensor = None` to a more explicit call to deallocate, i.e. `ttnn_tensor.deallocate()`, then by the time you call `ttnn.to_torch` the data is no longer valid and you will get a segmentation fault. Some careful considerations would needed to ensure the indirect references are not deallocated too early.

### What's changed

- Move the `TrackUnusedValues` class to its own pass. Generate calls to `.deallocate()` instead of assigning to `None`. 
  - As a result, the compiled code will also explicitly deallocate TTNN tensors now
  - Also, we will no longer need to insert code when exporting and instead grab the nodes corresponding to the `.deallocate()` calls. The exported code will match the compiled code more closely. 

Caveats:
- As mentioned above, indirect references can affect this pass. With the current set of models, I've seen `getitem` and `pack_to_tuples` being used. There could be other cases that I'm not aware of.
- Calling dead code elimination after this pass could also unintentionally remove the deallocation nodes since they are technically not "used" by others.
- The final code looks messier.

Most recent test run:
https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/14785670910


torch_ttnn/passes/deallocation_pass.py
```diff
@@ -0,0 +1,108 @@
...
+        # Instead we create a new list and iterate through that.
+        node_list = list(graph.nodes)
+        for node in node_list:
+            if (to_delete_nodes := unused_values.get_nodes_to_delete(node)) is not None:
+                for n in to_delete_nodes:
+                    # Skip nodes that are references to other nodes
+                    # We don't want to delete these too early
+                    if node.target in [target_wrappers.pack_to_tuple, operator.getitem]:
+                        continue
+                    with graph.inserting_after(node):
```
dgomezTT, line 99: quick question, what would happen if the node is used by to_torch? Does it make sense to keep it in memory until the end just for that?

dgomezTT (approved):

torch_ttnn/passes/deallocation_pass.py
kevinwuTT: Right now we're not moving any nodes around, so indeed, nodes that are used by to_torch directly won't be deallocated until close to the end. We definitely should consider this possibility in the future though.

philei-tt (approved):
Seems good to me

Issue #989, opened 2025-04-29
Lowering for `aten.native_layer_norm.default` does not match the number of outputs


kevinwuTT:
According to the reference, `aten.native_layer_norm.default` returns 3 outputs: layer norm output, mean, and rstd. [link](https://github.com/pytorch/pytorch/blob/f84062f78d723543e4962ffa1d38dcf42947e3f1/torch/_refs/__init__.py#L3321)

Currently, the lowering to `ttnn.layer_norm` only returns the layer norm output. This isn't typically an issue when running inferencing, but for training, the other two outputs are usually passed to the back prop function for `aten.native_layer_norm_backward` op.

Some affected models:
- SegFormer-train
- MLPMixer-train

PR #1004, opened 2025-05-06
Lower aten.native_layer_norm to ttnn.moreh.layer_norm

kevinwuTT:
### Ticket
https://github.com/tenstorrent/pytorch2.0_ttnn/issues/989

### Problem description
`torch.ops.aten.native_layer_norm.default` outputs a tuple of 3 tensors, but `ttnn.layer_norm` only outputs 1. Typically in inference, the other two tensors are unused. But for training, this will cause problems.

### What's changed
- Switch to use `ttnn.operations.moreh.layer_norm` which outputs 3 tensors. Since there's some additional setup, use a torch wrapper. 
- Add a test to check aten.native_layer_norm directly
- Add a test to check training version of torch.nn.LayerNorm

### Caveats
- moreh layer norm does not keep the normalized dim for the mean and rstd outputs. For example, for an input with shape (2, 32, 512) normalized on dim -1, aten layer norm will output the shape of (2, 32, 1), but moreh layer norm will give (2, 32). That's why there are reshapes at the end of this wrapper.
- moreh layer norm seems to only support inplace updating of the mean and rstd outputs. For example: [test_moreh_layer_norm.py](https://github.com/tenstorrent/tt-metal/blob/2aceab13b8e3ccfe9450e75dd5915d5be7893b44/tests/ttnn/unit_tests/operations/test_moreh_layer_norm.py#L99-L105)
- The dtype metadata for the mean and rstd outputs in the original aten graph do not exist. For example, the printout of `layer_norm_node.meta["val"]` looks like: 
  ```
  (FakeTensor(..., size=(2, 32, 64), dtype=torch.bfloat16), FakeTensor(..., size=(2, 32, 1)), FakeTensor(..., size=(2, 32, 1)))
  ```
  The data movement pass looks for the dtype of each tensor when generating `to_torch` ops. If they do not exist, it defaults to `float`. To fix this, the metadata for the mean and rstd nodes are copied from the input during the lowering stage.

Related issues:
- https://github.com/tenstorrent/tt-metal/issues/22089
- https://github.com/tenstorrent/tt-metal/issues/22110

### Questions
- [x] The special handling of `ttnn.layer_norm` in the data movement pass still remains. The moreh layer norm fully replaces `ttnn.layer_norm` in this compiler, but should the special handling be kept for compatibility purposes?


jmalone-tt:
Adding a note here about action items from meeting so I don't forget:

- [ ] Check performance to see if `moreh_layer_norm` is slower than `layer_norm`
- [ ] If perf is comparable, remove special cases for `layer_norm` and always lower to moreh version
- [ ] If perf is different, lower only when needed (look at users of outputs to determine)
- [ ] Talk to team in charge of `layer_norm` to understand if we should file ticket to have all 3 outputs or if we should just use `moreh_layer_norm` exclusively going forward

torch_ttnn/passes/lowering/target_wrappers.py
```diff
@@ -192,3 +192,39 @@ def shard_tensor(tensor, dim, num_devices):
...
+    ttnn_dtype: ttnn.DataType,
+    norm_dims: int,
+    gamma: ttnn.Tensor,
+    beta: ttnn.Tensor,
+    epsilon: ttnn.Tensor,
+    device: ttnn.Device,
+):
+    output = ttnn.empty(in_tensor_shape, device=device, layout=ttnn.TILE_LAYOUT, dtype=ttnn_dtype)
+
+    # BUG? moreh.layer_norm produces NaNs if the initial shape is the torch shape
```
ayerofieiev-tt, line 214: can you please fire a tt-metal ticket?

kevinwuTT:
@jmalone-tt 

See this code snippet for the implementation: https://gist.github.com/kevinwuTT/4ce5a0b50238ff148ceea3745facb2c9
In short, if we want to return all three outputs, `moreh.layer_norm` is slightly faster than computing all three separately using a combination of `ttnn` ops. On the other hand, if we only want the layer norm output, `ttnn.layer_norm` is slightly faster. 

However, we could detect if the mean and rstd outputs are used in a particular graph, and if they are not, then we can tell the `layer_norm` wrapper to call the `ttnn.layer_norm` op instead. 

In both scenarios, we can remove the special cases in the DataMovement pass because `ttnn.layer_norm` will always be under the wrapper, which will always return 3 outputs. (Mean and rstd will just return None if not needed.)

---
I've spoken to Borys about `ttnn.layer_norm` and he said there is currently no plan to support 3 outputs since it is meant to resemble `torch.nn.LayerNorm` and not `aten.native_layer_norm`. For our use case, `moreh.layer_norm` should be suitable. He did note that bug-fixes or changes to Moreh ops will be slower than core TTNN ops.

torch_ttnn/passes/lowering/target_wrappers.py
kevinwuTT: Done! I added the ticket to the PR description. https://github.com/tenstorrent/tt-metal/issues/22110

kevinwuTT:
Most recent successful test run: https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/15030679650

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -545,10 +525,7 @@ def align(self, node, input_node, input_site, input_site_type: InputSiteType, fi
                     aligned_node = self._create_aligned_node(data_move_spec)
             self.aligned_node_dict[data_move_spec] = aligned_node
 
-        if node.target == ttnn.layer_norm:
-            self._connect_aligned_node_layer_norm(node, input_node, aligned_node, input_site, input_site_type)
-        else:
-            self._connect_aligned_node(node, aligned_node, input_site, input_site_type)
+        self._connect_aligned_node(node, aligned_node, input_site, input_site_type)
```
jmalone-tt, line 528: Thanks for getting rid of special cases here!

torch_ttnn/passes/lowering/to_tt_pass.py
```diff
@@ -528,16 +526,60 @@ def cast_bf16(target, other):
...
+                use_rstd = False
+                # check the 2nd item if it's mean or rstd
+                if len(node_users) > 1:
+                    getitem = node_users[1]
+                    if getitem.args[1] == 1:
+                        use_mean = True
+                    if getitem.args[1] == 2:
+                        use_rstd = True
+                # if the 3rd item exists, it will always be rstd
+                if len(node_users) > 2:
```
jmalone-tt, line 557: Is this a reliable way to tell which outputs are accessed? I would imagine if we had a situation like the following:
```
a = layer_norm(inputs)
b = a + a
c = a * a
```
then we would see `node_users = [b,c]` even though only the output is used, not the mean or rstd. I could also be misunderstanding here

jmalone-tt (commented):
Thanks for checking with the team in charge of layer_norm and checking performance. A couple of small comments, mostly for my understanding, then it looks good to me!

torch_ttnn/passes/lowering/to_tt_pass.py
kevinwuTT: This is a good question, but I think this will work because ops like layer_norm always seem to follow with getitems. So the direct users of layer_norm will be either 1, 2, or 3. For example:

```
a = layer_norm(inputs)
getitem_0 = a[0]
b = getitem_0 + getitem_0
c = getitem_0 * getitem_0
```
So in this case, the `node_users` for `a` will only be `[getitem_0]` or a length of 1. But users for `getitem_0` will be 2 like in your example. 

I updated the comments to mention this.

jmalone-tt: Thanks for clarifying!

PR #1006, opened 2025-05-07
native cpp extension with cmake

philei-tt:
### Ticket
Original PR: https://github.com/tenstorrent/pytorch2.0_ttnn/pull/806 

### Problem description
This branch is based on https://github.com/tenstorrent/pytorch2.0_ttnn/tree/kw/cpp_extension 

### What's changed
On top of points mentioned in https://github.com/tenstorrent/pytorch2.0_ttnn/pull/806 
- Add instructions to compile torch to make it compatible with ttnn
- Using cmake to link torch and ttnn
- Optional git submodule to compile ttnn inside this repo
- Restructure C++ source files
- Fix a few minor bugs (lack of stides in TtnnTensorImpl, lack of ttnn->cpu copy)
- Add .clang-format
- Update docs with instructions on how to make native extension

Ideally we won't be recompiling torch, but compile ttnn in a compatible with torch variant (gcc + c++17 + some other flags used in torch), but that will be the next PR


torch_ttnn/cpp_extension/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+        ${TT_METAL_HOME}/tt_metal/api
+        ${TT_METAL_HOME}/tt_metal/third_party/umd/device/api
+        ${TT_METAL_HOME}/tt_metal/hostdevcommon/api
+        ${TT_METAL_HOME}/tt_metal/third_party/tracy/public
+        ${TT_METAL_HOME}/tt_stl
+        ${TT_METAL_HOME}/.cpmcache/reflect/e75434c4c5f669e4a74e4d84e0a30d7249c1e66f
+        ${TT_METAL_HOME}/.cpmcache/fmt/69912fb6b71fcb1f7e5deca191a2bb4748c4e7b6//include
+        ${TT_METAL_HOME}/.cpmcache/magic_enum/4d76fe0a5b27a0e62d6c15976d02b33c54207096/include
+        ${TT_METAL_HOME}/.cpmcache/nlohmann_json/798e0374658476027d9723eeb67a262d0f3c8308/include
+    )
```
kevinwuTT, line 54: On my 22.04 Ubuntu machine, I had to add `${TT_METAL_HOME}/.cpmcache/boost/1359e136761ab2d10afa1c4e21086c8d824735cd/libs/core/include`

torch_ttnn/cpp_extension/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+message(STATUS "Ubuntu version > 24: ${IS_UBUNTU_GE_24}")
+message(STATUS "CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
+
+# NOTE: ttnn reauires to downrade python version to 3.10, which break find_package
+# TODO: Remove this once ttnn is updated to support python 3.12+
+if (IS_UBUNTU_GE_24)
+    include(find_python_workaround_ubuntu24)
+else()
+    set(Python_FIND_VIRTUALENV FIRST)
+    find_package(Python COMPONENTS Interpreter Development REQUIRED)
```
kevinwuTT, line 25: For Ubuntu 22.04 I have to change Python to PythonLibs

tests/cpp_extension/test_cpp_extension_functionality.py
```diff
@@ -0,0 +1,112 @@
...
+    ((32, 1, 3, 3), (1, 32), (4,)),
+)
+@pytest.mark.parametrize(
+    "dtype",
+    (torch.bfloat16, torch.int, torch.long),
+)
+def test_cpp_extension(device, input_shape, dtype):
+    # in pytest the device has already been initialized before this call
+    # so instead we can wrap this around the custom device
+    ttnn_torch_device = ttnn_module.as_torch_device(device)
```
kevinwuTT, line 20: I'm getting this error:
```
E       TypeError: as_torch_device(): incompatible function arguments. The following argument types are supported:
E           1. (arg0: tt::tt_metal::IDevice) -> torch.device
E       
E       Invoked with: MeshDevice(1x2 grid, 2 devices)
```
It seems like my current setup is creating the device with `device = ttnn.open_device` instead of `device = ttnn.open_mesh_device`. I think we should still support both?

kevinwuTT (changes requested):
I haven't done a thorough review yet, but this first set of comments include some of the changes I had to make to get the cpp extension compiled.

docs/OpenRegistrationAPI.md
```diff
@@ -0,0 +1,110 @@
...
+export TT_METAL_HOME=/path/to/tt-metal
+cd torch_ttnn/cpp_extension
+python3 setup.py develop
+```
+
+2. Build fetch tt-metal from git submodule and build directly
+```bash
+git submodule sync
+git submodule update --init --recursive
+CMAKE_FLAGS="-DENABLE_SUBMODULE_TT_METAL_BUILD=ON" python3 setup.py develop
```
kevinwuTT, line 79: We should rewrite the steps when following the submodule path since we don't need to build metal separately, but we still need to create a new environment and build pytorch first. Or can we actually build pytorch as a third-party and metal at the same time with this path?

docs/OpenRegistrationAPI.md
```diff
@@ -0,0 +1,110 @@
...
+pip install -U numpy==1.26.4
+pip install -r requirements.txt
+
+CC=clang-17 CXX=clang++-17  CMAKE_POLICY_VERSION_MINIMUM=3.5 python setup.py develop
+```
+
+## Install pytorch2.0_ttnn dependencies
+NOTE: The ttnn packages are intentionally removed for this branch. They will be restored once a ttnn dev package containing all the dependencies is available.
+
+0. If you built torch from source, remove `torch` and `torchvision` from `requirements.txt`
```
kevinwuTT, line 52: We need to figure out a way to make sure locally built torch doesn't get overwritten. Some other packages like `mlp-mixer-pytorch` in requirements-dev.txt also install torch again.

Same for ttnn.

docs/OpenRegistrationAPI.md
```diff
@@ -0,0 +1,110 @@
...
+
+Run in the root of `pytorch2.0_ttnn` repo:
+```
+python -m pytest tests/cpp_extension/test_cpp_extension_functionality.py -s
+```
+
+You can also run with `DEBUG_CPP_EXT=1` ENV to enable logging messages.
+
+## Troubleshooting
+* You may have to set `PYTHONPATH=$TT_METAL_HOME:<location-of-pytorch2.0_ttnn-repo>` after all
```
jmalone-tt, line 107: Is there a certain error message / condition that would indicate `PYTHONPATH` needs to be set? If so, we should mention when this is applicable

tests/cpp_extension/test_bert_cpp_extension.py
```diff
@@ -0,0 +1,117 @@
...
+        'hellenist who first articulated the difference between greek, greco - roman and roman art. " the prophet and founding hero of modern archaeology ", winckelmann was one of the founders of scientific archaeology',
+    ]
+
+    inputs = [
+        tokenizer.encode_plus(
+            q,
+            context,
+            add_special_tokens=True,
+            return_tensors="pt",
+            max_length=256,
```
jmalone-tt, line 42: Another PR changes max_length to 384. I am guessing this doesn't impact correctness, but it does keep us in line with metal for performance testing.

tests/cpp_extension/test_cpp_extension_functionality.py
```diff
@@ -0,0 +1,112 @@
...
+    input_b = torch.randint(-1000, 1000, input_shape, dtype=dtype)
+
+    logging.info("Transferring to ttnn")
+    input_a_ttnn = input_a.to(ttnn_torch_device)
+    input_b_ttnn = input_b.to(ttnn_torch_device)
+
+    logging.info("Get underlying ttnn tensor")
+    ttnn_tensor_a = ttnn_module.get_ttnn_tensor(input_a_ttnn)
+    ttnn_tensor_b = ttnn_module.get_ttnn_tensor(input_b_ttnn)
+    # ttnn_tensor_scalar = ttnn_module.get_ttnn_tensor(scalar_ttnn)
```
jmalone-tt, line 81: delete?

tests/cpp_extension/test_cpp_extension_functionality.py
```diff
@@ -0,0 +1,112 @@
...
+
+    # Compare output of abs op for bfloat16 dtype since ttnn.abs does not support int
```
jmalone-tt, line 83: delete / change comment

tests/cpp_extension/test_cpp_extension_functionality.py
```diff
@@ -0,0 +1,112 @@
...
+    logging.info("Creating bfloat tensor from -1 to 1")
+    torch_tensor = torch.empty((32, 1, 3, 3), dtype=torch.bfloat16, device=ttnn_torch_device)
+
+    logging.info("Get underlying ttnn tensor")
+    ttnn_tensor = ttnn_module.get_ttnn_tensor(torch_tensor)
+
+    logging.info("calling to_torch")
+    ttnn_to_torch = ttnn.to_torch(ttnn_tensor)
+
+    # assert torch.allclose(torch_tensor, ttnn_to_torch, rtol=0.1, atol=0.1)
```
jmalone-tt, line 112: Is there a reason this is commented out?

torch_ttnn/cpp_extension/ttnn_cpp_extension/include/ttnn_cpp_extension/core/TtnnCustomAllocator.hpp
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include <c10/core/Allocator.h>
+
+// A dummy allocator used for creating Storage for Torch Tensors
+// This should not use any actual space (data is stored in DRAM, not CPU RAM)
```
jmalone-tt, line 6: Do we need to handle host tensors here? Or are they also in DRAM?

torch_ttnn/cpp_extension/ttnn_cpp_extension/src/core/TtnnGuard.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+at::Stream TtnnDeviceGuard::exchangeStream(at::Stream) const noexcept {
+    // no-op
+    return at::Stream(at::Stream::DEFAULT, at::Device(at::DeviceType::PrivateUse1, CURR_DEVICE));
+}
+
+at::DeviceIndex TtnnDeviceGuard::deviceCount() const noexcept {
+    // Hardcoding the number of "valid" devices here at 2.
+    return 2;
```
jmalone-tt, line 63: Do we have a plan to get this number programmatically in the future? Hardcoding to 2 doesn't seem sustainable long term

torch_ttnn/cpp_extension/ttnn_cpp_extension/src/core/copy.cpp
```diff
@@ -0,0 +1,120 @@
...
+            // Create TTNN Tensor on CPU
+            ttnn::Tensor src_cpu = ttnn::Tensor(
+                tt::tt_metal::HostStorage{tt::tt_metal::host_buffer::create<bfloat16>(
+                    tt::stl::Span(self_bfloat16_storage_ptr, logical_volume))},
+                logical_shape,
+                dtype,
+                ttnn::Layout::ROW_MAJOR);
+
+            // Verify the data is created on TTNN tensor correctly
+            // compare_torch_and_ttnn_tensors(self, src_cpu);
```
jmalone-tt, line 46: delete or uncomment these checks? Also some below

torch_ttnn/cpp_extension/ttnn_cpp_extension/src/core/copy.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+            // TODO: Find out why there are problems when passing device directly to `to_layout` function
+            ttnn::Tensor src_dev = src_cpu.to_device(ttnn_device);
+            ttnn::Tensor res = ttnn::to_layout(src_dev, ttnn::TILE_LAYOUT, std::nullopt, std::nullopt, ttnn_device);
```
jmalone-tt, line 50: Is it faster to tilize on device than CPU? Guessing that's the motivation for keeping the `ttnn::to_layout` call separate, but it would be good to have a comment giving the reason

torch_ttnn/cpp_extension/ttnn_cpp_extension/src/utils/vector_utils.cpp
```diff
@@ -0,0 +1,85 @@
...
+            auto torch_vector = std::get<std::vector<float>>(tensor_to_vector(torch_tensor));
+            auto ttnn_vector = std::get<std::vector<float>>(tensor_to_vector(ttnn_tensor));
+
+            vector_compare(torch_vector, ttnn_vector);
+        } else if (ttnn_dtype == ttnn::DataType::UINT32) {
+            auto torch_vector = std::get<std::vector<int>>(tensor_to_vector(torch_tensor));
+            auto ttnn_vector = std::get<std::vector<int>>(tensor_to_vector(ttnn_tensor));
+
+            vector_compare(torch_vector, ttnn_vector);
+        }
```
jmalone-tt, line 83: Is it worth adding a guard for other types?

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -439,8 +440,18 @@ def _create_aligned_node(self, spec):
                 kwargs["layout"] = spec.layout()
             if spec.dtype is not None:
                 kwargs["dtype"] = spec.dtype()
-            return self.graph.call_function(ttnn.from_torch, (spec.input_node,), kwargs)
-
+            if (
+                "val" in spec.input_node.meta
+                and hasattr(spec.input_node.meta["val"], "device")
+                and str(spec.input_node.meta["val"].device) == "ttnn:0"
```
jmalone-tt, line 446: Guessing this is another point of incompatibility with mesh devices. Good to keep for now, but would be good to talk through how this change will need to be handled when there are multiple devices in play

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -439,8 +440,18 @@ def _create_aligned_node(self, spec):
...
+            ):
+                aligning_nodes = []
+                aligning_nodes.append(self.graph.call_function(ttnn_module.get_ttnn_tensor, (spec.input_node,), {}))
+                if node.target == ttnn.embedding and input_site == 0:
```
jmalone-tt, line 450: Why do we have special handling for embedding here? Seems like this should be handled by the pre-populated kwargs, either in the call to add `get_ttnn_tensor` or just a set of calls to match the populated kwargs for all node.targets. This would let us get rid of the `node` and `input_site` arguments again.

jmalone-tt (commented):
I haven't tested this change on my machine, but the code looks good overall. I left couple of comments and a couple of questions to make sure I'm understanding things

torch_ttnn/cpp_extension/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+        ${TT_METAL_HOME}/.cpmcache/nlohmann_json/798e0374658476027d9723eeb67a262d0f3c8308/include
+    )
+elseif(ENABLE_SUBMODULE_TT_METAL_BUILD)
+    add_subdirectory(third-party)
+else()
+    message(FATAL_ERROR "TT_Metal build is not enabled. Please set ENABLE_LOCAL_TT_METAL_BUILD to ON.")
+endif()
+
+
+set(TTNN_CPP_EXTENSION_SOURCES
```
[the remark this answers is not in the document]
philei-tt: Yes mistake. It is included below but under wrong if condition

torch_ttnn/cpp_extension/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+add_library(ttnn_device_extension SHARED ${TTNN_CPP_EXTENSION_SOURCES})
+target_compile_options(ttnn_device_extension PRIVATE
+    -DFMT_HEADER_ONLY
+    -DTORCH_EXTENSION_NAME=ttnn_device_extension
+    -DTORCH_API_INCLUDE_EXTENSION_H
+)
+target_include_directories(ttnn_device_extension PRIVATE 
+    ${CMAKE_CURRENT_SOURCE_DIR}/ttnn_cpp_extension/include
+    ${PYTHON_INCLUDE_DIRS}
+)
```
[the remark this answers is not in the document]
philei-tt: Yes mistake. It is included below but under wrong if condition

torch_ttnn/cpp_extension/CMakeLists.txt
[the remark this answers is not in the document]
philei-tt: I know, but I think it's better to explicitly see which files are used for compilation. For example in ttnn source code they also add all files manually. If you glob, there would be a chance to add unwanted file to compilation

docs/OpenRegistrationAPI.md
philei-tt: I think we can build pytorch as third-party, but then we will have to try to figure out how to install python part of it from cmake. I think it's kind of logical that you don't need to build tt-metal, but I will add a comment. But if you won't build tt-metal, you still need to use ttnn from wheel, or we should figure out how to install python part of ttnn from built submodule 🤔

docs/OpenRegistrationAPI.md
philei-tt: I think we can't really control what user does with their environment. That's one more reason to try to make ttnn abi compatible with torch and not other way around

tests/cpp_extension/test_bert_cpp_extension.py
philei-tt: Gotcha. Do I need to update inputs too?

tests/cpp_extension/test_cpp_extension_functionality.py
philei-tt: Thanks for attention to details, I always miss those things)

tests/cpp_extension/test_cpp_extension_functionality.py
philei-tt: No point in coparing vectors coming from empty(), since data is uninitialized. But I can compare shapes and dtype though

torch_ttnn/cpp_extension/ttnn_cpp_extension/include/ttnn_cpp_extension/core/TtnnCustomAllocator.hpp
philei-tt: We shouldn't: wrapper (TtnnTensorImpl) holds ttnn tensor as it's field and ttnn manages all the allocations, that why this allocator shouldn't be doing anything.

torch_ttnn/cpp_extension/ttnn_cpp_extension/src/core/TtnnGuard.cpp
philei-tt: Good spot, I think we can fetch it right now. I wanted to replace that eventually for sure

tests/cpp_extension/test_bert_cpp_extension.py
jmalone-tt: Nope. I should have a change up soon that moves questions and answers to a separate json file - we can reference that here or keep it the same. Changing `max_length` should just pad or truncate as needed.

torch_ttnn/cpp_extension/ttnn_cpp_extension/src/core/copy.cpp
philei-tt: I didn't verify if it is 🤔 . Should be? I will try to quickly benchmark later

torch_ttnn/cpp_extension/ttnn_cpp_extension/src/utils/vector_utils.cpp
philei-tt: Lol this one is not used anymore. It was used earlier for debugging purposes. I will remove it

torch_ttnn/passes/lowering/add_data_move_pass.py
philei-tt: 100%

torch_ttnn/passes/lowering/add_data_move_pass.py
philei-tt: @kevinwuTT

torch_ttnn/cpp_extension/CMakeLists.txt
philei-tt: Interesting. Maybe in recent tt-metal commits somebody exposed boost in api headers

torch_ttnn/cpp_extension/CMakeLists.txt
philei-tt: yes, I can confirm that, will update

docs/OpenRegistrationAPI.md
philei-tt: Yes I will add pip uninstall lower, to make sure that torch is not overwritten

docs/OpenRegistrationAPI.md
philei-tt: Updated

torch_ttnn/cpp_extension/ttnn_cpp_extension/src/core/copy.cpp
philei-tt: Deleted

tests/cpp_extension/test_cpp_extension_functionality.py
philei-tt: I can't reproduce that

docs/OpenRegistrationAPI.md
```diff
@@ -0,0 +1,110 @@
...
+    pip install -r requirements-dev.txt
+    ```
+
+    You can ignore the following message for now:
+    ```
+    ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
+    tt-smi 3.0.12 requires pre-commit==3.5.0, but you have pre-commit 3.0.4 which is incompatible.
+    ```
+
+## Building cpp extension
```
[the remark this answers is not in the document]
kevinwuTT: I updated the instructions to separate the two paths here: https://github.com/tenstorrent/pytorch2.0_ttnn/commit/21fdf9c5fd2d3c8bb5417e5296491cdeaf66d148

tests/cpp_extension/test_cpp_extension_functionality.py
kevinwuTT: Yeah I don't see this error anymore after the recent updates.

torch_ttnn/passes/lowering/add_data_move_pass.py
kevinwuTT: The original reason was because all of the tensors are tilized when they are preloaded. (torch_ttnn/cpp_extension/ttnn_cpp_extension/src/core/copy.cpp::ttnn_copy_from). At the same time, the input parameter for `ttnn.embedding` (site 0) only supports non-tiled uint32 tensors. So this special case was added to counter that. I agree that this is a bit messy, but I'm currently not sure how to individually assign the layouts for each input weight during the preloading stage.

docs/OpenRegistrationAPI.md
kevinwuTT: I rewrote the instructions here: https://github.com/tenstorrent/pytorch2.0_ttnn/blob/21fdf9c5fd2d3c8bb5417e5296491cdeaf66d148/docs/OpenRegistrationAPI.md. It doesn't seem like pytorch needs to be rebuild for the submodule path? Granted, I've only tested the simple unit test, and I haven't encountered any errors yet.

philei-tt: It works, but it's hard to be 100% sure that it won't break in the future. I think it will work as long as we guarantee that submodule version == ttnn wheel version. Ideally we should install ttnn as python package from submodule, but we can do that later in the future

torch_ttnn/cpp_extension/ttnn_cpp_extension/src/core/copy.cpp
philei-tt: Average <function HOST_TILIZER at 0x7f91645ebd90> time over 1000 runs: 0.003319 seconds
Average <function DEVICE_TILIZER at 0x7f9121e36cb0> time over 1000 runs: 0.000780 seconds

4.25 times quicker to move to device and then tilize

ayerofieiev-tt: @philei-tt are you aware about any cases which tilize on device can't handle?

docs/OpenRegistrationAPI.md
jmalone-tt: Can you add a change to this PR in `workflows/update-ttnn-wheel.yaml` to update the metal submodule commit to match the updated wheel release? This should automatically keep our wheel and metal submodule in sync.

jmalone-tt:
My only other remaining concern here is CI time. Do you have an idea of how this change will impact CI times with rebuilding? If you haven't tested, it would be good to run it through and we can discuss caching on NFS if needed.

torch_ttnn/cpp_extension/ttnn_cpp_extension/src/ops/binary.cpp
```diff
@@ -0,0 +1,53 @@
...
+    LOGGING("input device: {}, other device: {}", input.device(), other.device());
+    TORCH_CHECK(input.device().type() == c10::DeviceType::PrivateUse1);
+    TORCH_CHECK(other.device().type() == c10::DeviceType::PrivateUse1);
+
+    at::TtnnTensorImpl* tensor_impl = static_cast<at::TtnnTensorImpl*>(input.unsafeGetTensorImpl());
+    auto ttnn_tensor_input = tensor_impl->get_ttnn_tensor();
+    at::TtnnTensorImpl* other_tensor_impl = static_cast<at::TtnnTensorImpl*>(other.unsafeGetTensorImpl());
+    auto ttnn_tensor_other = other_tensor_impl->get_ttnn_tensor();
+
+    std::cout << "[ttnn_cpp_extension] ttnn_add_out" << std::endl;
```
kevinwuTT, line 23: I have the `LOGGING` macro. Should we use that?

kevinwuTT (approved):
LGTM! Just some minor comments.

torch_ttnn/cpp_extension/ttnn_cpp_extension/src/ops/binary.cpp
philei-tt: Oops, i wasn't supposed to push that

torch_ttnn/cpp_extension/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+option(BUILD_EXAMPLES "Build examples" OFF)
+
+list(PREPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+include(utils)
+
+check_ubuntu_version_at_least(24 IS_UBUNTU_GE_24)
+message(STATUS "Ubuntu version > 24: ${IS_UBUNTU_GE_24}")
+message(STATUS "CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
+
+# NOTE: ttnn reauires to downrade python version to 3.10, which break find_package
```
ayerofieiev-tt, line 19: ttnn must not force 3.10 on 24.04, the policy is that ttnn is supposed to use a system version (3.10 on 22.04 and higher on 24.04)

torch_ttnn/cpp_extension/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+        ${TT_METAL_HOME}/.cpmcache/reflect/e75434c4c5f669e4a74e4d84e0a30d7249c1e66f
+        ${TT_METAL_HOME}/.cpmcache/fmt/69912fb6b71fcb1f7e5deca191a2bb4748c4e7b6//include
+        ${TT_METAL_HOME}/.cpmcache/magic_enum/4d76fe0a5b27a0e62d6c15976d02b33c54207096/include
+        ${TT_METAL_HOME}/.cpmcache/nlohmann_json/798e0374658476027d9723eeb67a262d0f3c8308/include
+        ${TT_METAL_HOME}/.cpmcache/boost/1359e136761ab2d10afa1c4e21086c8d824735cd/libs/core/include
+    )
+elseif(ENABLE_SUBMODULE_TT_METAL_BUILD)
+    add_subdirectory(third-party)
+else()
+    message(FATAL_ERROR "TT_Metal build is not enabled. Please set ENABLE_LOCAL_TT_METAL_BUILD to ON.")
```
ayerofieiev-tt, line 59: ENABLE_SUBMODULE_TT_METAL_BUILD vs ENABLE_LOCAL_TT_METAL_BUILD whats the difference?

philei-tt: ENABLE_LOCAL_TT_METAL_BUILD -- you pass TT_METAL_HOME, to a tt-metal project you built locally with build_metal.sh, which is nice if you need to change ttnn

ENABLE_SUBMODULE_TT_METAL_BUILD - fetch and build ttnn as dependency directly in pytroch2 compiler, you can't really change it since version is hardcoded

torch_ttnn/cpp_extension/CMakeLists.txt
philei-tt: It must not but right now it doesn't work, since ttnn pushes you to use older pip version

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "torch_ttnn/cpp_extension/third-party/tt-metal"]
+	path = torch_ttnn/cpp_extension/third-party/tt-metal
+	url = https://github.com/tenstorrent/tt-metal
+	branch = v0.58.0-rc25
```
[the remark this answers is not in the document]
philei-tt: https://github.com/tenstorrent/pytorch2.0_ttnn/commit/12e27063ef75a218e7e89a25b2e87236c37b92aa

torch_ttnn/cpp_extension/CMakeLists.txt
philei-tt: Those options are removed now

tests/cpp_extension/test_mnist_cpp_extension.py
```diff
@@ -0,0 +1,64 @@
...
+        verbose=True,
+    )
+
+    model = MnistModel()
+    model = model.to(torch.bfloat16)
+    test_input = test_input.to(ttnn_device)
+    model.to(ttnn_device)
+
+    model = torch.compile(model, backend=torch_ttnn.backend, options=option)
+    results = model(test_input)
```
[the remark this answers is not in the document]
philei-tt: There is something like that in bert (https://github.com/tenstorrent/pytorch2.0_ttnn/blob/d09065df284d64d49dbb6f64c69d5559a3594a30/tests/cpp_extension/test_bert_cpp_extension.py#L90): 

iter 0: 29404.42503094673 (ms)
iter 1: 77.31628894805908 (ms)
iter 2: 76.47160172462463 (ms)
iter 3: 76.39674115180969 (ms)
iter 4: 76.4088921546936 (ms)

Mnist is skipped right now anyway. Not sure why, @kevinwuTT maybe you know?

docs/OpenRegistrationAPI.md
```diff
@@ -0,0 +1,168 @@
...
+2. Build torch
+    ```
+    CC=clang-17 CXX=clang++-17 CMAKE_POLICY_VERSION_MINIMUM=3.5 python setup.py develop
+
+    popd
+    ```
+
+### Build torchvision (Optional)
+Optionally, you can build torchvision, which might be required for some computer-vision models:
+    ```bash
```
kevinwuTT, line 131: I think we need to remove the indents, because it looks like this in the preview:
Screenshot 2025-05-20 at 11 51 52 AM

philei-tt: Good catch

tests/cpp_extension/test_mnist_cpp_extension.py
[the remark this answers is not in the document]
kevinwuTT: @philei-tt Initially there were compatibility issues with convolution. Could be layout related.

.github/workflows/community-issue-tagging.yml
```diff
@@ -1,4 +1,4 @@
-name: "Community Issue / PR Labeling Workflow"
+name: Community Issue / PR Labeling Workflow
```
ayerofieiev-tt, line 1: why "" removed here while new `.github/workflows/run-cpp-native-tests.yaml` has "" ?

philei-tt: Oops messed up rebase conflict, I will revert to what is on main

jmalone-tt (approved):

PR #1015, opened 2025-05-14
Load Weights Once

jmalone-tt:
This adds changes to only load weights once by adding a wrapper function with logic to load weights on device and use cached tensors for future iterations.

Cached tensors should be deallocated when the calling function ends. Since Native Device should have the same behavior, this is ok for now. If we start running into issues with running out of DRAM, we may need to be more strategic about which tensors are cached on device.

Note that this PR does not enable cached model parameters for training runs, only inference. This is accomplished through an additional GraphModuleAnalysisPass. This pass looks for characteristics that are typical of backward computation graphs (calls to backward aten ops), forward training graphs (outputting one of the inputs unchanged), and marks everything else as forward inference. This seems fragile, but any of the normal torch methods of determining whether a model is in training or eval mode appear to not work by the time the GraphModule gets to our backend.

This PR also fixes an issue where inputs for data parallel were not hoisted to the top of the forward function.

This PR also opens up the opportunity for preprocessing model weights. This is needed for convolution ops to be able to run metal trace, and will help speed up several other functions once implemented. Preprocessing will be addressed in a separate PR.

Local tests indicate the following performance improvements:
MNIST Batch 1, 5th iteration changes **5.29 ms -> 3.32 ms**
BERT Batch 8, 5th iteration now hits **23.3 sen/sec** (batch 1 on main today hits ~1.7 sen/sec, not sure about batch 8)


jmalone-tt:
Based on a quick test, this lowers 5th iteration MNIST batch 1 from 5.29 ms to 3.32 ms. I have not tested BERT, data parallel, or other models.

torch_ttnn/backend.py
```diff
@@ -187,7 +192,8 @@ def aten_backend(
     pm = PassManager(passes=passes)
     gm, modified = pm(gm)
 
-    GraphCleanup(gm)
+    # GraphCleanup(gm)
```
dgomezTT, line 195: any reason to leave the comment?

torch_ttnn/passes/analysis/graph_module_analysis_pass.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class GraphModuleAnalysisPass(PassBase):
+    def __init__(self):
+        super().__init__()
+
+    def call(self, gm: torch.fx.GraphModule):
+        """Marks inputs for the compiled function as parameters, buffers, or arguments.
+
+        This relies on how PyTorch behaves under the hood. Namely, PyTorch will lay out function arguments in the order of parameters, then buffers, then arguments. We can rely on this layout order and the number of parameters, buffers, and arguments passed in from AOTAutograd to our backend function.
```
dgomezTT, line 61: this is great, it helps a lot to understand the logic behind it!

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -553,6 +620,46 @@ def align(self, node, input_node, input_site, input_site_type: InputSiteType, fi
         return 1
 
 
+def insert_load_params_once(gm, first_node, nodes, node_input_aligner):
+    SiteType = NodeInputAligner.InputSiteType
+    i = 0
+
+    for node in nodes:
+        args = node.args
```
kevinwuTT, line 628: I know args and kwargs are iterated differently, but the rest of the code seems like it could be combined. Let me know if it's more difficult than it looks.

kevinwuTT (approved):
LGTM! I'm glad the analysis pass for determining training/eval works well enough.

jmalone-tt:
Passed tests:
https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/15048228625

torch_ttnn/passes/lowering/add_data_move_pass.py
jmalone-tt: I'm not entirely sure what change you're asking for here. Can you clarify?

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -270,6 +271,8 @@ def __init__(self, graph, device):
         self.graph = graph
         self.device = device
         self.aligned_node_dict = {}
```
ayerofieiev-tt, line 273: Can you please leave a comment on what this member is about?

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -529,11 +579,28 @@ def align(self, node, input_node, input_site, input_site_type: InputSiteType, fi
 
         if data_move_spec in self.aligned_node_dict:
             aligned_node = self.aligned_node_dict[data_move_spec]
+        elif ttnn_inputs is not None and data_move_spec in self.marshaled_node_dict:
+            # add getitem call
+            # update aligned_node_dict
+            input_idx = self.marshaled_node_dict[data_move_spec]
```
ayerofieiev-tt, line 585: Here you are looking in dict twice. In above if statement with IN and now with access. I assume it’s not a big deal, but want to highlight in case it matters

ayerofieiev-tt (approved):
In general looks alright. Maybe a tiny bit hacky, but that’s the only way I can see right now too.

jmalone-tt:
This passes consistently in the `Run Tests` workflow, and passes locally, but fails consistently in `Before Merge` 😕. Need to investigate further.

torch_ttnn/passes/analysis/graph_module_analysis_pass.py
```diff
@@ -0,0 +1,78 @@
...
+    It is expected that PARAMETER and BUFFER tensors do not change between inference runs, but ARGUMENT tensors may change.
+
+    :param PARAMETER: Tensor tracked by optimizer during training. Represents weights and biases.
+    :param BUFFER: Tensor not updated during training but still part of model state. Normally used for non-trainable data like fixed weights or running statistics in batch norm.
+    :param ARGUMENT: Tensors not tracked by model and not persistent between calls. Can be input data or configuration options to module's methods.
+    """
+
+    INFERENCE = 1
+    TRAIN_FORWARD = 2
+    TRAIN_BACKWARD = 3
```
philei-tt, line 21: I think docs are from other enum? From PrimalTag?

torch_ttnn/passes/analysis/graph_module_analysis_pass.py
```diff
@@ -0,0 +1,78 @@
...
+
+aten_backward_ops = {
+    torch.ops.aten._adaptive_avg_pool2d_backward.default,
+    torch.ops.aten.avg_pool2d_backward.default,
+    torch.ops.aten.convolution_backward.default,
+    torch.ops.aten.embedding_dense_backward.default,
+    torch.ops.aten.max_pool2d_with_indices_backward.default,
+    torch.ops.aten.native_group_norm_backward.default,
+    torch.ops.aten.native_layer_norm_backward.default,
+}
```
philei-tt, line 32: Are you sure this list will always determine that graph is train backward? Feels too small)

torch_ttnn/passes/analysis/graph_module_analysis_pass.py
```diff
@@ -0,0 +1,78 @@
...
+def is_train_backward(gm):
+    node_list = list(gm.graph.nodes)
+    for node in node_list:
+        if node.op == "call_function" and node.target in aten_backward_ops:
+            return True
+
+    return False
+
+
+def is_train_forward(gm):
```
philei-tt, line 44: Is this documented in somewhere in torch docs that train forward will return inputs? If no, this might change in the future

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -520,7 +541,39 @@ def _connect_aligned_node_layer_norm(
...
+            # No need to align input_node
+            return 0
+
+        if data_move_spec in self.marshaled_node_dict:
+            # already handled
+            return 0
+
+        with self.graph.inserting_before(first_node):
+            self.marshaled_node_dict[data_move_spec] = self.input_idx
+            self.input_idx += 1
```
philei-tt, line 572: Why is this block wrapped with with self.graph.inserting_before? You aren't inserting node here, or am I missing something?

torch_ttnn/passes/lowering/target_wrappers.py
```diff
@@ -6,6 +6,25 @@
 
 from torch_ttnn.utils import TtnnDevice
 
+run_once_count = 0
+run_once_ans = tuple()
+
+
+@torch.fx.wrap
+def run_once(*args):
```
philei-tt, line 14: (nit) using closure or @lru_cache might be a bit better then global variable.

torch_ttnn/passes/lowering/add_data_move_pass.py
```diff
@@ -573,26 +681,34 @@ def call(self, gm: torch.fx.GraphModule):
...
                     for tuple_idx, tuple_arg in enumerate(arg):
                         i += node_input_aligner.align(
-                            node, tuple_arg, [key, tuple_idx], SiteType.KWARGS_TUPLE, first_node
+                            node, tuple_arg, [key, tuple_idx], SiteType.KWARGS_TUPLE, first_node, ttnn_inputs
                         )
                 else:
-                    i += node_input_aligner.align(node, arg, key, SiteType.KWARGS, first_node)
+                    i += node_input_aligner.align(node, arg, key, SiteType.KWARGS, first_node, ttnn_inputs)
 
         modified = i > 0
```
philei-tt, line 713: (nit) maybe use modified = modified or node_input_aligner.align(...) for readability purposes? Not sure it's better, but variable "i" doesn't tell much about it's purpose

philei-tt (approved):
Looks good, just a few small comments.

torch_ttnn/passes/analysis/graph_module_analysis_pass.py
jmalone-tt: Thanks for catching this!

torch_ttnn/passes/analysis/graph_module_analysis_pass.py
jmalone-tt: I don't know if it's documented anywhere, but inputs are used in backward pass, so this should be reliable. Can add a comment that notes this assumption though

torch_ttnn/passes/lowering/add_data_move_pass.py
jmalone-tt: Good point, earlier in the change I was adding the nodes here, but then removed that code. Will update

torch_ttnn/passes/lowering/target_wrappers.py
jmalone-tt: I really like the lru_cache here, but it hits an error with the kwargs inputs being unhashable since it's a dict. The closure path seems less readable to me in this case. I agree that the global variable isn't the best, but I'm leaning towards keeping it unless you feel strongly

philei-tt: of course, it's no a big issue

torch_ttnn/passes/analysis/graph_module_analysis_pass.py
philei-tt: To be honest, I'm not convinced that all models will have aten.*_backward.* calls in backward pass. Here is an example that produces forward+backward aten IR for small linear model, that doesn't have any functions from the list: 

```python
import torch
import torch.nn as nn
from torch.fx import symbolic_trace
from torch.fx.experimental.proxy_tensor import make_fx

# Define a simple model with 2 linear layers and ReLU
class SimpleModel(nn.Module):
    def __init__(self, in_features=10, hidden_features=5, out_features=2):
        super(SimpleModel, self).__init__()
        self.fc1 = nn.Linear(in_features, hidden_features)
        self.relu = nn.ReLU()
        self.fc2 = nn.Linear(hidden_features, out_features)

    def forward(self, x):
        print(f"Input shape: {x.shape}")
        x = self.fc1(x)
        print(f"After fc1: {x.shape}")
        x = self.relu(x)
        print(f"After relu: {x.shape}")
        x = self.fc2(x)
        print(f"After fc2: {x.shape}")
        return x

if __name__ == "__main__":
    # Instantiate model and dummy input
    model = SimpleModel()
    x = torch.randn(1, 10, requires_grad=True)

    # Forward pass with shape printing
    y = model(x)

    # Trace the forward graph with torch.fx
    traced = symbolic_trace(model)
    print("\n--- Forward FX Graph ---")
    print(traced.graph)
    print("\n--- Forward FX Graph (Python IR) ---")
    print(traced.code)

    # Define a function combining forward & backward
    def forward_and_backward(inp):
        out = model(inp)
        loss = out.sum()
        loss.backward()
        return out

    # Use make_fx to capture both forward and backward aten IR
    fx_g = make_fx(forward_and_backward)(x)
    print("\n--- Forward+Backward FX Graph (aten IR) ---")
    print(fx_g.graph)
    
    print("\n --- Forward+Backward FX Graph (Python IR) ---")
    print(fx_g.code)
```

Forward + backward generated python code:
```
 --- Forward+Backward FX Graph (Python IR) ---



def forward(self, inp_1):
    _param_constant0 = self._param_constant0
    t = torch.ops.aten.t.default(_param_constant0);  _param_constant0 = None
    _param_constant1 = self._param_constant1
    addmm = torch.ops.aten.addmm.default(_param_constant1, inp_1, t);  _param_constant1 = None
    relu = torch.ops.aten.relu.default(addmm);  addmm = None
    detach = torch.ops.aten.detach.default(relu)
    _param_constant2 = self._param_constant2
    t_1 = torch.ops.aten.t.default(_param_constant2);  _param_constant2 = None
    _param_constant3 = self._param_constant3
    addmm_1 = torch.ops.aten.addmm.default(_param_constant3, relu, t_1);  _param_constant3 = None
    sum_1 = torch.ops.aten.sum.default(addmm_1)
    ones_like = torch.ops.aten.ones_like.default(sum_1, pin_memory = False, memory_format = torch.preserve_format);  sum_1 = None
    expand = torch.ops.aten.expand.default(ones_like, [1, 2]);  ones_like = None
    t_2 = torch.ops.aten.t.default(t_1);  t_1 = None
    mm = torch.ops.aten.mm.default(expand, t_2);  t_2 = None
    t_3 = torch.ops.aten.t.default(expand)
    mm_1 = torch.ops.aten.mm.default(t_3, relu);  t_3 = relu = None
    t_4 = torch.ops.aten.t.default(mm_1);  mm_1 = None
    sum_2 = torch.ops.aten.sum.dim_IntList(expand, [0], True);  expand = None
    view = torch.ops.aten.view.default(sum_2, [2]);  sum_2 = None
    detach_1 = torch.ops.aten.detach.default(view);  view = None
    detach_2 = torch.ops.aten.detach.default(detach_1);  detach_1 = None
    t_5 = torch.ops.aten.t.default(t_4);  t_4 = None
    detach_3 = torch.ops.aten.detach.default(t_5);  t_5 = None
    detach_4 = torch.ops.aten.detach.default(detach_3);  detach_3 = None
    detach_5 = torch.ops.aten.detach.default(detach);  detach = None
    threshold_backward = torch.ops.aten.threshold_backward.default(mm, detach_5, 0);  mm = detach_5 = None
    t_6 = torch.ops.aten.t.default(t);  t = None
    mm_2 = torch.ops.aten.mm.default(threshold_backward, t_6);  t_6 = None
    t_7 = torch.ops.aten.t.default(threshold_backward)
    mm_3 = torch.ops.aten.mm.default(t_7, inp_1);  t_7 = inp_1 = None
    t_8 = torch.ops.aten.t.default(mm_3);  mm_3 = None
    sum_3 = torch.ops.aten.sum.dim_IntList(threshold_backward, [0], True);  threshold_backward = None
    view_1 = torch.ops.aten.view.default(sum_3, [5]);  sum_3 = None
    detach_6 = torch.ops.aten.detach.default(view_1);  view_1 = None
    detach_7 = torch.ops.aten.detach.default(detach_6);  detach_6 = None
    detach_8 = torch.ops.aten.detach.default(mm_2);  mm_2 = None
    detach_9 = torch.ops.aten.detach.default(detach_8);  detach_8 = None
    t_9 = torch.ops.aten.t.default(t_8);  t_8 = None
    detach_10 = torch.ops.aten.detach.default(t_9);  t_9 = None
    detach_11 = torch.ops.aten.detach.default(detach_10);  detach_10 = None
    return addmm_1
 ```

jmalone-tt: Ok, looks like that graph contains `torch.ops.aten.threshold_backward.default`, which isn't in the list. I will go through and check the models that we support and make sure each of the backward passes would be caught in practice

Issue #992, opened 2025-04-30
Unify operations that can be resolved by higher level ops


dgomezTT:
A good example of the linear operation for bert, or the softmax whose patterns are repeated all over the compiled code and can be unified under a ttnn op, like linear or attention_softmax_

New tickets will be created for future unifications

The goal of this is to improve performance

PR #1024, opened 2025-05-19
Fuse operations in Bert attention

dgomezTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/pytorch2.0_ttnn/issues/992)

### Problem description
There are some operations in the compiled code that can be unified in a higher level call.
Bert attention should work with concatenated heads and reduce the amount of low level operations

### What's changed
Added a new compilation pass that unifies bert attention layers with higher level operations.

Before PR:
Runtime Iteration without cache: 9.11285948753357
Runtime Iteration with cache: 0.38252902030944824

After PR:
Runtime Iteration without cache: 8.339576005935669
Runtime Iteration with cache: 0.37560153007507324

Generated code here:
https://github.com/dgomezTT/BertCompiledOptimization/commit/1b31451af6c13ffaf66816bea95524afe67baad3

Before:
Runtime Iteration 0: 7.632845163345337
Runtime Iteration 1: 0.3204987049102783
Runtime Iteration 2: 0.30744123458862305
Runtime Iteration 3: 0.30896854400634766
Runtime Iteration 4: 0.3109018802642822

After:
Runtime Iteration 0: 7.582945108413696
Runtime Iteration 1: 0.3196885585784912
Runtime Iteration 2: 0.30839991569519043
Runtime Iteration 3: 0.3023970127105713
Runtime Iteration 4: 0.3016688823699951

One layer comparison and full code comparison:
https://github.com/dgomezTT/BertCompiledOptimization/commit/028db647fd771222d83f0915ddd3ea0e73869242

Hardware used:
n150d + ryzen 9 5950x3d

### DIFF
--- a/torch_ttnn/passes/patterns/attention_pattern.py
+++ b/torch_ttnn/passes/patterns/attention_pattern.py
@@ -16,12 +16,13 @@ class AttentionPatterns(PatternMatcherBase):
     def match_pattern(self) -> List[Tuple[torch.fx.Node, ...]]:
         """
         Match the attention pattern:
-        linear -> view -> reshape -> permute -> view -> matmul(q, k) -> view -> softmax -> view -> matmul(softmax, v)
+        linear -> view -> reshape -> permute -> view -> matmul(q, k) -> view -> softmax -> view -> matmul(softmax, v) -> view -> view -> permute -> reshape -> view
 
         Returns a list of tuples: (q_linear, q_view1, q_reshape, q_permute, q_view2,
                                  k_linear, k_view1, k_reshape, k_permute, k_transpose, k_view2,
                                  v_linear, v_view1, v_reshape, v_permute, v_view2,
-                                 matmul_qk, view1, softmax, view2, matmul_sv, attn_mask, scale)
+                                 matmul_qk, view1, softmax, view2, matmul_sv, view3, permute, reshape, view5,
+                                 attn_mask, scale)
         """
         matches = []
 
@@ -72,6 +73,23 @@ class AttentionPatterns(PatternMatcherBase):
                     if not matmul_sv or matmul_sv is not v_matmul:
                         continue
 
+                    # Check for additional operations after matmul_sv
+                    view3 = self._find_exclusive_user_of_type(matmul_sv, ttnn.experimental.view)
+                    if not view3:
+                        continue
+
+                    permute = self._find_exclusive_user_of_type(view3, ttnn.permute)
+                    if not permute:
+                        continue
+
+                    reshape = self._find_exclusive_user_of_type(permute, ttnn.reshape)
+                    if not reshape:
+                        continue
+
+                    view5 = self._find_exclusive_user_of_type(reshape, ttnn.experimental.view)
+                    if not view5:
+                        continue
+
                     # We found a complete attention pattern
                     matches.append(
                         (
@@ -96,12 +114,38 @@ class AttentionPatterns(PatternMatcherBase):
                             softmax,
                             view2,
                             matmul_sv,
+                            view3,
+                            permute,
+                            reshape,
+                            view5,
                             attn_mask,
                             scale,
                         )
                     )
 
         return matches
+
+    # Hack! TTNN does not support concat with tensors of 1D
+    # https://github.com/tenstorrent/tt-metal/issues/20205
+    # bias is 1D tensor, so we need to make it 2D
+    def _hack_1D_bias(self, bias_node, node_before_bias):
+        """Reshape bias node to 2D Shape
+
+        Args:
+            bias_node: The bias node to reshape
+
+        Returns:
+            The reshaped node
+        """
+
+        with self.gm.graph.inserting_after(node_before_bias):
+            new_bias_node = self.gm.graph.call_function(
+                ttnn.experimental.view,
+                args=(bias_node, (1, -1)),
+                kwargs={},
+            )
+
+        return new_bias_node
 
     def replace_pattern(self, matches: List[Tuple[torch.fx.Node, ...]]) -> None:
         for match in matches:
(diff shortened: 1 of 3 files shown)


torch_ttnn/passes/patterns/attention_pattern.py
```diff
@@ -0,0 +1,422 @@
...
+                qvk_node = self.gm.graph.call_function(
+                    ttnn.linear,
+                    args=(
+                        q_linear.args[0],
+                        concatenated_qkv_weights_node,
+                    ),
+                    kwargs={"bias": concatenated_qkv_biases_node},
+                )
+
+            with self.gm.graph.inserting_after(qvk_node):
```
jmalone-tt, line 244: Looks like there is the pattern here:
```python
with self.gm.graph.inserting_after(parent_node):
    new_node = self.gm.graph.call_function(...)

with self.gm.graph.inserting_after(new_node):
    another_new_node = self.gm.graph.call_function(...)

...
```

It reads more cleanly if you just use `self.gm.graph.inserting_before(node_you_want_after_all_inserted_functions)`
^ you can think of it like a linked list, and this repeatedly inserts just before whichever node you pass in, so you can call `self.gm.graph.call_function` several times and the order you call it will match the resulting order in the computation graph afterwards

jmalone-tt: In case my explanation isn't clear,

```python
with self.gm.graph.inserting_before(node_d):
    self.gm.graph.call_function(func_a)
    b = self.gm.graph.call_function(func_b)
    self.gm.graph.call_function(func_c, (b))
```

will end up like:

```
%node_a = ttnn_decorators_ttnn_func_a()
%node_b = ttnn_decorators_ttnn_func_b()
%node_c = ttnn_decorators_ttnn_func_c (%node_b)
%node_d = ttnn_decorators_ttnn_func_d()
```

torch_ttnn/passes/patterns/attention_pattern.py
```diff
@@ -0,0 +1,422 @@
...
+            with self.gm.graph.inserting_after(linear_view_node):
+                # Time to split the qvk into q, k, v
+                split_qkv_node = self.gm.graph.call_function(
+                    ttnn.transformer.split_query_key_value_and_split_heads,
+                    args=(linear_view_node, None),
+                    kwargs={"num_heads": num_heads, "transpose_key": True},
+                )
+
+            mask_new_shape = list(attn_mask.meta["val"].shape)
+            mask_new_shape[2], mask_new_shape[3] = mask_new_shape[3], mask_new_shape[2]
```
jmalone-tt, line 263: Dead code?

torch_ttnn/passes/patterns/attention_pattern.py
```diff
@@ -0,0 +1,422 @@
...
+
+            with self.gm.graph.inserting_after(attention_probabilities_node):
+                matmul_sv_node = self.gm.graph.call_function(
+                    ttnn.matmul,
+                    args=(attention_probabilities_node, get_v_node),
+                    kwargs={},
+                )
+
+            with self.gm.graph.inserting_after(matmul_sv_node):
+                concate_head_node = self.gm.graph.call_function(
```
jmalone-tt, line 308: nit: spelling

torch_ttnn/passes/patterns/attention_pattern.py
```diff
@@ -0,0 +1,422 @@
...
+                k_view1,
+                v_view1,
+                q_linear,
+                k_linear,
+                v_linear,
+            ]
+            self.safe_remove_nodes(nodes_to_remove)
+
+        self.gm.graph.lint()
+        self.gm.recompile()
```
jmalone-tt, line 347: Fine to leave in here, but I think GraphCleanup runs after all the passes, which should call these functions too

torch_ttnn/passes/patterns/attention_pattern.py
```diff
@@ -0,0 +1,422 @@
...
+        Args:
+            linear: The linear node to start from
+            require_transpose: Whether to require a transpose operation
+            matmul_arg_idx: Which argument of matmul should be the view2 node (0 for query, 1 for key/value)
+
+        Returns:
+            Tuple of nodes if pattern matches, None otherwise
+        """
+
+        # Even though the bias is optional, I want to restrict the pattern to only include nodes with a bias
```
jmalone-tt, line 362: Will this change at some point, or will we only ever convert nodes with bias?

jmalone-tt (changes requested):
Nicely done getting this working!
Would also be good to have a test for the pattern replacement so we can ensure the pattern is being matched and swapped

torch_ttnn/passes/patterns/attention_pattern.py
```diff
@@ -0,0 +1,422 @@
...
+        """Reshape bias node to 2D Shape
+
+        Args:
+            bias_node: The bias node to reshape
+
+        Returns:
+            The reshaped node
+        """
+
+        with self.gm.graph.inserting_after(node_before_bias):
```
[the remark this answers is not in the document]
dgomezTT: this is kinda hacky. The idea is to reshape nodes that are 1D to 2D until that bug is fixed in metal. Ideally we will get rid of all of this functionality

torch_ttnn/passes/patterns/attention_pattern.py
dgomezTT: mmm, yes, that would do it. I can definitely refactor it

torch_ttnn/passes/patterns/attention_pattern.py
dgomezTT: yes, this was my hack to make sdpa work. The need a mask (b, 1, s, s) so I did a reshape and a ttnn.add of a mask (b, 1, s, 1) and (b, 1, 1, s)
Good catch!

torch_ttnn/passes/patterns/attention_pattern.py
dgomezTT: my bad, it was my early bug detection mechanism. I'll get rid of it!

torch_ttnn/passes/patterns/attention_pattern.py
dgomezTT: I did not want to make this overly generic, considering the implications of changing the compiler. Since the pattern for bert which is our target, requires a bias, I thought about adding that as a restriction and controlling a bit the spread of this change. I am happy to go all in and get rid of this limitation if you think is worth it

torch_ttnn/passes/patterns/attention_pattern.py
```diff
@@ -0,0 +1,413 @@
...
+                        kwargs={},
+                    )
+
+                with self.gm.graph.inserting_after(transposed_v_weights_node):
+                    # Lets concatenate the qkv tensors for weights and biases
+                    concatenated_qkv_weights_node = self.gm.graph.call_function(
+                        ttnn.concat,
+                        args=([transposed_q_weights_node, transposed_k_weights_node, transposed_v_weights_node], -1),
+                        kwargs={},
+                    )
```
kevinwuTT, line 216: If I understand correctly, these also should not need the `inserting_after` either.

kevinwuTT (changes requested):
LGTM, just a few comments.

torch_ttnn/passes/patterns/attention_pattern.py
dgomezTT: good catch!

kevinwuTT (approved):
Thanks for the changes. LGTM

jmalone-tt (approved):

dgomezTT:
https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/15165949024

PR #1050, opened 2025-05-22
Move preprocessing into `run_once`

jmalone-tt:
This PR is the first part of preprocessing convs so we can support metal trace and have even better perf.

In particular, it:
* Removes the `run_once` logic from AddDataMovePass
* Adds another pass called LoadOncePass that handles loading and preprocessing weights
* Adds support for arbitrary functions in the `run_once` wrapper
    * This is pretty hacky. We can't directly pass in functions, so we pass a fully qualified function name as a string, then decode it in the run_once wrapper function
    * We try to reduce variable liveness by deallocating temporaries in `run_once`. Previously calculated values are referenced by their index inside `run_once`, which is captured by `RunOnceIdx`. To pass to the function, we must serialize, and here we use pickle
*  Adds more functions to the list of backward functions based on our current op tracking

Once this PR is merged, I will add handling for preprocessing convs, since this is a bit more involved


ayerofieiev-tt (approved):

torch_ttnn/backend.py
```diff
@@ -190,7 +193,11 @@ def aten_backend(
         passes = passes_with_graphviz
 
     pm = PassManager(passes=passes)
-    gm, modified = pm(gm)
+    try:
+        gm, modified = pm(gm)
+    except Exception as e:
+        logging.error(e.__context__)
```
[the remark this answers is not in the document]
jmalone-tt: Good question. Right now, if there is an error during one of the passes, we just get a message about which pass the error occurred in. This line prints additional context, like what the error actually was. It spits out more text when there's an error, but has helped me diagnose without having to add a breakpoint in the exception handler of the PassManager's `__call__` method.

torch_ttnn/passes/analysis/graph_module_analysis_pass.py
```diff
@@ -19,16 +19,27 @@ class ModelType(Enum):
     TRAIN_BACKWARD = 3
 
 
-# this list seems small, but is based off all the backward calls from the Core Aten IR:
+# This list seems small, but is based off all the backward calls from the Core Aten IR:
 # https://docs.pytorch.org/docs/stable/torch.compiler_ir.html
+# The list is augmented by any ops we track in docs/operations, which helps capture more in practice
 aten_backward_ops = {
```
philei-tt, line 25: What do you think about changing this to something like: 
```python
def is_train_backward(gm):
    node_list = list(gm.graph.nodes)
    for node in node_list:
      if (
          isinstance(node.target, str)
          and "backward" in node.target
          and node.op == "call_function"
      ):
        return True

    return False
```

jmalone-tt: Updated to guard against node.op first, then check if backward is in target name

jmalone-tt:
This change makes 4 models fail, 3 of which were previous E2E. Marked compilation_xfail for now

PR #1054, opened 2025-05-23
Fix flaky lowering tests and speed up CI

jmalone-tt:
There are a couple of small changes bundled together in this PR:
* Fix tanh and pow_scalar tests by manually seeding torch.rand
    * These have been flaky, which causes unnecessary retriggering of before merge
* Fix bug in Update Documentation flow that prevented actually merging doc updates
* Fix how inputs are passed to Run Tests from Before Merge
    * This bug caused all model tests to be run for 5 iterations, which slows down Before Merge significantly

### DIFF
--- a/tests/lowering/eltwise/binary/test_pow_scalar.py
+++ b/tests/lowering/eltwise/binary/test_pow_scalar.py
@@ -8,8 +8,6 @@ import ttnn
 
 from tests.utils import assert_with_pcc
 
-torch.manual_seed(1337)
-
 
 class PowTensorScalarModule(torch.nn.Module):
     def __init__(self):
@@ -40,4 +38,4 @@ def test_pow_tensor_scalar_square(device, input, exponent_shape):
     assert [node.target for node in nodes].count(ttnn.pow) == 1
     assert result_before.shape == result_after.shape
     # Check inference result
-    assert_with_pcc(result_before, result_after)
+    assert_with_pcc(result_before, result_after, pcc=0.997)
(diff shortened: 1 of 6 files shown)


tests/lowering/eltwise/binary/test_pow_scalar.py
```diff
@@ -8,6 +8,8 @@
 
 from tests.utils import assert_with_pcc
 
+torch.manual_seed(1337)
```
dgomezTT, line 11: 1337 sounds really specific. Does this number has a meaning? should it be a constant?

tests/lowering/eltwise/unary/test_tanh.py
```diff
@@ -8,6 +8,8 @@
 
 from tests.utils import assert_with_pcc
 
+torch.manual_seed(1337)
```
dgomezTT, line 11: how many tests will need this number? should the base test class take care of it?

dgomezTT (approved):
looks good to me, I guess that 1337 might have a good explanation, but it would be amazing if we document it better!

tests/lowering/eltwise/unary/test_tanh.py
jmalone-tt: That's why the tests are flaky now. If you don't manually seed, then `torch.randn` will get random values. `tanh` usually passes, but sometimes barely misses the PCC threshold

tests/lowering/eltwise/binary/test_pow_scalar.py
philei-tt: https://en.wikipedia.org/wiki/Leet

tests/lowering/eltwise/unary/test_tanh.py
philei-tt: Yeah tests should have seed. But also, if you see that they sometimes fail for random seed, then maybe it's better to lower threshold too?

philei-tt (approved):

PR #1069, opened 2025-05-27
Preprocess conv weight and bias

jmalone-tt:
Updates compiler to preprocess convolution weight and bias in `run_once` function
Note that the current merge target is `more_preprocessing` to show diff more cleanly, but it will switch to `main` once the other PR is merged.
Additionally, this PR marks many models as compilation_xfail, most of which hit errors with L1 OOM


jmalone-tt:
This change makes 8 models fail, 1 of which was previously E2E. Marked compilation_xfail for now, but will re-enable in future.
This regression is acceptable because the average speedup for E2E models is 20x

torch_ttnn/passes/lowering/load_once_pass.py
```diff
@@ -80,6 +87,61 @@ def move(self, node):
...
         return 1
 
+    def move_conv(self, node):
+        new_args = list(node.args)
+        modified_count = 0
+
+        # rewrite input because otherwise it is not defined when run_once is called
+        input_node = node.args[0]
+        tensor_shape = list(input_node.meta["val"].shape)
+        spec_dtype = TtnnBfloat16
```
dgomezTT, line 97: can you get that info from the arguments?

dgomezTT: considering that you evaluate that later, you can ignore that comment!

dgomezTT (approved):

torch_ttnn/passes/lowering/load_once_pass.py
```diff
@@ -80,6 +87,61 @@ def move(self, node):
...
+        if get_dtype(input_node) in [torch.int32, torch.int64]:
+            spec_dtype = TtnnUint32
+        # currently, all conv inputs are tiled and on device. If this changes, may need further analysis here
+        mocked_input_node = (
+            False,
+            "ttnn.ones",
+            (),
+            {"shape": tensor_shape, "device": TtnnDevice(), "dtype": spec_dtype(), "layout": TtnnTileLayout()},
```
dgomezTT, line 105: is TtnnUint32 or any of them a function?

torch_ttnn/passes/lowering/load_once_pass.py
jmalone-tt: These are good notes - I can add a comment to the code to explain, but at this point, the `val` stored in the argument nodes are all torch tensors. We could add another analysis pass that marks all nodes with their dtype, but this seemed quicker in the short term. If there are other passes that need this information, it will probably be worth adding a pass though.

jmalone-tt:
Looks like tests pass with recent commit, so we do not lose any models as a result of this change!
https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/15349376232

torch_ttnn/passes/lowering/target_wrappers.py
```diff
@@ -37,10 +38,10 @@ def lookup_function(str_name):
                 base = getattr(base, part)
             return base
 
-        def lookup_prev_result(maybe_pickled_run_once_idx):
+        def lookup_prev_result_idx(maybe_pickled_run_once_idx):
```
kevinwuTT, line 41: I think it would be helpful to make a comment on the reasoning behind needing to convert this idx to a byte object and loading it here.

kevinwuTT (approved):
Overall LGTM!

PR #1078, opened 2025-05-29
Average compiled runtime over several runs

jmalone-tt:
Recently, we have seen lots of variability in compiled model runtime. This may be due to only reporting the 5th iteration instead of averaging over several iterations.

This PR changes the reported number to be averaged over any iterations after the first (creates program cache, much slower) and the second (may still be warming up caches). The new metric is reported as `avg_run_time` in the model metrics.

In the future, it may also be good to report TTFT in the front page (basically the first iteration, but we might need to report when the output begins streaming if it takes a while to return multiple values)


kevinwuTT (approved):
Thanks. LGTM.

PR #1080, opened 2025-05-30
Match tt-metal batch sizes for testing models

jmalone-tt:
Changing batch size can greatly impact both throughput and TTFT for models.
This PR:
* Adds the ability to set batch_size in `_load_inputs` for each model
* Matches the batch sizes listed for tt-metal's front page as closely as possible (sometimes we need to choose a lower batch size to not hit an OOM)
* Adds batch size and throughput to the README for reporting performance, as well as an approximation of TTFT

Note that all models have batch_size=1 in this PR except ResNet50 and BERT (4 and 8 respectively). This PR just attempts to match batch size for models on the tt-metal front page. Increasing batch sizes for other models will be left for another PR.

### DIFF
--- a/tools/collect_metrics.py
+++ b/tools/collect_metrics.py
@@ -34,13 +34,17 @@ csv_header_mappings = {
         "To/From Device Ops",
         "The number of `to/from_device` operations (data transfer to/from the device).",
     ),
-    "original_run_time": (
-        "Original Run Time (ms)",
-        "Execution time (in seconds) of the model before conversion.",
+    "original_throughput": (
+        "Original Throughput (Inferences Per Second)",
+        "Execution throughput (in inferences per second) of the model before conversion.",
     ),
-    "compiled_run_time": (
-        "Compiled Run Time for 5th Iteration (ms)",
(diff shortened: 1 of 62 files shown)


ayerofieiev-tt (approved):

tools/collect_metrics.py
```diff
@@ -481,15 +489,19 @@ def serialize_to_pydantic_operations(self):
...
         pydantic_model = pydantic_models.ModelRun(name=model)
 
+        batch_size = compiled_runtime_metrics.get("batch_size", 1)
+
         # Rename run_time keys to distinguish between original or compiled
+        pydantic_model.batch_size = batch_size
         if "run_time" in original_runtime_metrics:
             run_time = original_runtime_metrics.pop("run_time")
-            original_runtime_metrics["original_run_time"] = run_time
+            original_runtime_metrics["original_throughput"] = batch_size / run_time * 1000
```
kevinwuTT, line 498: Should we consolidate the units? The "run_time" that's calculated in conftest.py is in milliseconds.

kevinwuTT (approved):
LGTM, just a comment about standardizing the units of measurement.

PR #1089, opened 2025-06-02
Workflows - update-readme: Add permissions that seem to be missing

warthog9:
Pretty sure the token involved here doesn't have the ability to grant `gh` the power to do the approval as it is, I believe this sorts it.

### Ticket
Link to Github Issue

### Problem description

### What's changed


jmalone-tt:
Fixed in another PR that also adds permissions to other workflows

PR #1103, opened 2025-06-09
Update permissions for failing workflows

jmalone-tt:
Recent change to enterprise GH was causing many workflows to fail. This has been fixed at the enterprise level, but we still have a couple workflows that do not request the correct permissions. This PR should fix those workflows


.github/workflows/run-accuracy-tests.yaml
```diff
@@ -72,7 +73,6 @@ jobs:
         with:
           splits: ${{ needs.count-test-files.outputs.num_files }}
           matrix_group: ${{ matrix.group }}
-          commit_report: ${{ github.event.inputs.commit_report }}
```
dgomezTT, line 75: any reason to do this?

dgomezTT (approved):

.github/workflows/run-accuracy-tests.yaml
jmalone-tt: Admittedly a bit outside the scope of permissions, but this workflow doesn't actually take `commit_report` as an input, so it was always set to None anyway

kevinwuTT (approved):
LGTM

PR #1106, opened 2025-06-10
Do not use in-place attention softmax operation

jmalone-tt:
This fixes the issue where calling deallocate didn't actually deallocate the softmax tensor since we were previously using the in-place operation

The previous bug caused an issue for L1 memory configs since memory wasn't actually deallocated when tensor.deallocate() was called. This PR changes the op so memory will be deallocated


kevinwuTT (approved):
LGTM!

dgomezTT (approved):
great catch!

Issue #1072, opened 2025-05-28
Collect and upload tracy perf reports during stats update


ayerofieiev-tt:
Today we collect performance stats like - end-to-end model call duration.
The task is to collect and upload tracy perf reports to be readily available alongside latest metrics.

PR #1116, opened 2025-06-17
Generate tracy artifacts for end-to-end models

kevinwuTT:
## Ticket
[Link to Github Issue](https://github.com/tenstorrent/pytorch2.0_ttnn/issues/1072)

### Overview:
- Add a separate Github action that builds tt-metal
  - [x] Combine this with native device to avoid duplicate code
- Integrate a tt-metal build into test runs if requested
- Insert tracy instrumentation into lowering with a pass
  - Also insert into conftest

### Workflow process:
1. If "Docs" or "All" is called, it will compile tt-metal with tracy enabled. It will copy the build into NFS.
1. For each matrix group:
    1. Copy the cached build from NFS to current work directory.
    1. Separate subtests by using pytest flags `--collect-only`. This allows tracy produce data for each subtest instead of combining them into one. 
        1. In conftest, only end-to-end converted models will be profiled. Otherwise, the test runs normally. 
    1. Upload the `tracy_out` data

### Additional changes:
- Moved tt-metal submodule into tools/third-party instead of cpp_extension.
- Add some additional flags to to build tt-metal with tracy or torch compatibility

### Remaining TODOs:
- [ ] Make sure all end-to-end tests pass or add additional markings for tests failing with tracy

### DIFF
--- a/tests/models/torchvision/test_torchvision_image_classification.py
+++ b/tests/models/torchvision/test_torchvision_image_classification.py
@@ -46,7 +46,7 @@ model_info_and_mode_list = [
     ),
     pytest.param(
         [("densenet169", "DenseNet169_Weights"), "eval"],
-        marks=pytest.mark.converted_end_to_end,
+        marks=[pytest.mark.converted_end_to_end, pytest.mark.tracy_incompatible],
     ),
     pytest.param(
         [("densenet201", "DenseNet201_Weights"), "eval"],
--- a/.github/actions/common_model_tests/action.yaml
+++ b/.github/actions/common_model_tests/action.yaml
@@ -42,6 +42,7 @@ runs:
           fi
         }
 
+        # Setup tracy dependencies
         if [ "$run_tracy" = true ];
         then
           echo "Copying tt-metal build from NFS cache ($CACHE_DIR)."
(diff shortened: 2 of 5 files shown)


torch_ttnn/passes/tracy_profiling_pass.py
```diff
@@ -0,0 +1,30 @@
...
+        self.enable_tracy_profiling = enable_tracy_profiling
+
+    def call(self, gm: torch.fx.GraphModule):
+        if self.enable_tracy_profiling:
+            graph = gm.graph
+
+            num_nodes = len(graph.nodes)
+            for i, node in enumerate(graph.nodes):
+                # Insert ttnn.DumpDeviceProfiler every 500 nodes and before the last
+                if i % 500 == 0 or i == (num_nodes - 1):
```
philei-tt, line 24: Why 500? Is it a random number, or is it taken from somewhere?

kevinwuTT: According to the docs here: https://github.com/tenstorrent/tt-metal/blob/main/docs/source/ttnn/ttnn/profiling_ttnn_operations.rst, it's recommended to call this every 1000 ops to avoid dropping profiling data. If I can recall, I don't think it hurts to reduce this so that's why I chose 500. But I can probably increase this back to 1000 if it produces stable results.

tests/models/torchvision/test_torchvision_image_classification.py
```diff
@@ -40,7 +40,10 @@ def _load_inputs(self, batch_size):
 model_info_and_mode_list = [
     [("googlenet", "GoogLeNet_Weights"), "eval"],
     pytest.param([("densenet121", "DenseNet121_Weights"), "eval"], marks=pytest.mark.converted_end_to_end),
-    pytest.param([("densenet161", "DenseNet161_Weights"), "eval"], marks=pytest.mark.converted_end_to_end),
+    pytest.param(
+        [("densenet161", "DenseNet161_Weights"), "eval"],
+        marks=[pytest.mark.converted_end_to_end, pytest.mark.skip(reason="Failing in tracy")],
```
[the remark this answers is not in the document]
kevinwuTT: Yeah you're right; this will unconditionally skip the test. However, I should be able to define a new marker just for tracy. Shouldn't be difficult and I can add it to this PR.

torch_ttnn/passes/tracy_profiling_pass.py
philei-tt: Gotcha. (nit) better to name this number to avoid magical numbers. Also "If profiling data is dropped, you will receive warning messages in the execution log "

PR #1136, opened 2025-06-30
Migrate multi device tests to CIv2

jmalone-tt:
Recently, we have had issues with several GHA runners being flaky. In particular, the N300 runners have been down, which has blocked merges.

This PR adds support for running the multi-device tests in CIv2 infrastructure. It is also the first step to adding CIv2 support to all of our actions.

The main changes include modifying how the multi-device tests use pip and fixing a recursion bug in today's implementation of the multi-device shard analysis pass.

### DIFF
--- a/.github/actions/common_cleanup/action.yaml
+++ b/.github/actions/common_cleanup/action.yaml
@@ -6,6 +6,7 @@ runs:
     - name: Cleanup model cache
       shell: bash
       run: |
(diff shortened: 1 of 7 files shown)


jmalone-tt:
Tests passed:
https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/15984596252

PR #1159, opened 2025-07-10
Cache eager model test results

jmalone-tt:
Today, we rerun the eager mode of each model every time it is run for tests, but we expect these results to not change from run to run.

By caching, we can reduce CI time considerably. This also sets us up to be better citizens for CIv2 usage.

For example, this reduces BERT batch 8 model test from 17:06 -> 00:45 (though I expect most model tests will not see as significant of a speedup)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,7 +8,7 @@ import ttnn
 import torch
 import torch_ttnn
 import collections
-from tests.utils import calculate_accuracy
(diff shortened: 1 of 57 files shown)


kevinwuTT (approved):
I think this looks good overall. I left one comment about the saving of the eager results. 

Just one additional thought: Should we have a flag to skip reading from cache and/or writing to cache? I was thinking we could have a workflow that performs a full reset of the cache if needed. But this also be a separate PR.

PR #1175, opened 2025-07-23
Align pip package with recent tt-metal release

jmalone-tt:
Removes [ttnn] optional install, now installs ttnn by default
Updates ttnn version to match what is latest on PyPI

In the future, we should automatically check for releases and update to the latest one. I was playing around with this today, but it gets a bit annoying if we want [dev] to point to the latest nightly


jmalone-tt:
Currently testing here:
https://github.com/tenstorrent/pytorch2.0_ttnn/actions/runs/16480777341

PR #1180, opened 2025-08-05
Add Gaussian Splatting model

kevinwuTT:
## Problem description
Imported gaussian splatting implementation written by @hbb1 from https://github.com/hbb1/torch-splatting.

## Changes from original implementation
All files under the `tests/models/gaussian_splatting/gaussian_splatting` directory are adapted from the original repo above.
* Removed hardcoded CUDA references
* Add CPU version of computing KNN distances
* Modified some parts of the GSSTrainer class to be more flexible when passing configuration values around
* Added TTNN device and Pytorch-TTNN compile statements
* Added @torch.no_grad() decorator for eval step to reduce memory usage
* Add support for FP16. For the CPU run, during training, the RAM usage can go above 20GB, which can cause the Linux OOM-killer to end the process early. This has been observed during CI runs.
  * Below is an mprof plot generated from a local run on a system with about 47GB of RAM using float32:
    plot
  * Below is using float16. Peak memory usage is about 13GB.
    plot

Many of the changes above were also initially co-authored by jbedichekTT 
### In pytorch2.0_ttnn
* Ported the changes made in `train.py` from torch-splatting repo to pytest
* This test will download a snapshot of the tt-compat branch of the torch-splatting fork, extract to a directory, and then include the directory into `PYTHONPATH` so the modules become available
* Test run log `python -m pytest -s tests/models/gaussian_splatting`:
  * [log.log](https://github.com/user-attachments/files/21587373/log.log)
* Accuracy, the strikethrough was using FP32 and current changes use FP16 for CPU
  * train: ~~0.6721~~ 0.4200 
  * eval: ~~0.3576~~ 0.5014
* Skipped MLP-Mixer because of OOM issues https://github.com/tenstorrent/pytorch2.0_ttnn/issues/1184


tests/models/gaussian_splatting/test_gaussian_splatting.py
```diff
@@ -3,14 +3,16 @@
...
+import requests
+import subprocess
 import sys
+import torch
+import zipfile
+
+from tests.utils import ModelTester
 
 # Path to patched gaussian splatting implementation that is compatible with TT
 github_url = "https://github.com/kevinwuTT/torch-splatting/archive/refs/heads/tt-compat.zip"
```
ayerofieiev-tt, line 18: whats the patch? :) how to reproduce?

kevinwuTT: This is the diff https://github.com/kevinwuTT/torch-splatting/commit/2c95025d59dfcd9d4e92625b2655de1b87526cf2. 

You should be able to run this test by simply running: `python -m pytest -s tests/models/gaussian_splatting`. We can discuss what is the best way to handle an external module. I'm open to suggestions.

tests/models/gaussian_splatting/test_gaussian_splatting.py
```diff
@@ -0,0 +1,161 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+# Reference: https://huggingface.co/docs/transformers/v4.44.2/en/model_doc/gpt_neo#overview
```
jmalone-tt, line 4: Can delete this line

tests/models/gaussian_splatting/test_gaussian_splatting.py
```diff
@@ -43,6 +45,10 @@ def setup_gs_module():
     with zipfile.ZipFile(f"{gs_module_path}/B075X65R3X.zip", "r") as zip_ref:
```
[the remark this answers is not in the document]
jmalone-tt: Maybe adding comments in the code giving explanations of the dataset (what it consists of, where it is from) would be a good idea

tests/models/gaussian_splatting/test_gaussian_splatting.py
```diff
@@ -0,0 +1,161 @@
...
+        response = requests.get(github_url)
+    except Exception as e:
+        print(f"Error downloading URL: {github_url}")
+        raise e
+
+    assert response.status_code == 200
+
+    # Write to disk and extract
+    zip_path = f"{test_path}/tt-compat.zip"
+    open(zip_path, "wb").write(response.content)
```
jmalone-tt, line 37: I think the zip file makes sense for training data, but I'm wondering if it would be better to either migrate the TT-compliant model definition inside PT2.0 or add it as a git submodule.
1. Would either of these be a difficult change?
2. Do you have opinions here (mostly concerned about long-term maintenance)?

kevinwuTT: 1. It looks like we only need around 6 files, so importing these wouldn't be difficult at all. 
1. I think having the model definition in PT2.0 might be it easier to maintain, as long as we're allowed to. I don't think I see any particular licenses that prohibits us from from porting most of the code to this repo. But let me know otherwise.

